PR #16, opened 2023-10-21
statistics counts mindless mob kills

xDanilcusx:
it is what it is


morrowwolf:
I'm good to add this for now just if people start getting weird about kill totals I'm gonna be sad and revert this. I hope you don't take any offense either way 🙏

xDanilcusx:
It's okay, hope this one will not cause any problems

PR #17, opened 2023-10-21
Facehugger AI & autodoc now can extract embryos

xDanilcusx:
### but you can have a look at AI behaviour in action [here](https://discord.com/channels/1160427683151368192/1160428588043075625/1196614771143876678)


morrowwolf:
Will TM and review this coming Monday.

code/modules/mob/living/carbon/xenomorph/ai/movement/facehugger.dm
```diff
@@ -0,0 +1,130 @@
...
+
+/datum/xeno_ai_movement/linger/facehugger/ai_move_target(delta_time)
+	var/mob/living/carbon/xenomorph/moving_xeno = parent
+
+	if(moving_xeno.action_busy)
+		return
+
+	return ..()
+
+#define REENGAGE_COOLDOWN (7 SECONDS)
```
morrowwolf, line 51: These should be converted to variables now that they're used in multiple places so we can just override as needed

code/modules/mob/living/carbon/xenomorph/ai/movement/facehugger.dm
```diff
@@ -0,0 +1,130 @@
...
+		return
+
+	if(!travelling_turf || get_dist(travelling_turf, moving_xeno) <= 0)
+		for(var/i = 0 to FIND_NEW_TRAVEL_TURF_LIMIT)
+			travelling_turf = get_random_turf_in_range_unblocked(moving_xeno.current_target, linger_range + linger_deviation, linger_range - linger_deviation)
+
+			if(!travelling_turf)
+				continue
+
+			if(get_dist(travelling_turf, moving_xeno.current_target) <= 4)
```
morrowwolf, line 69: If this is the only change just make a new var or function that gets overrided, we don't want to copy code if we can help it so people in the future don't have to change code in multiple places.

code/modules/mob/living/carbon/xenomorph/castes/Facehugger.dm
```diff
@@ -19,17 +19,17 @@
 	caste_type = XENO_CASTE_FACEHUGGER
 	speak_emote = list("hisses")
 	icon_state = "Facehugger"
-	icon_size = 48
+	icon_size = 45
```
morrowwolf, line 22: Is this intentional? I'm not sure what this would effect

code/modules/mob/living/carbon/xenomorph/castes/Facehugger.dm
```diff
@@ -5,9 +5,9 @@
 	plasma_max = 10
 	melee_damage_lower = 5
 	melee_damage_upper = 5
-	max_health = XENO_HEALTH_LARVA
+	max_health = 45
```
morrowwolf, line 8: Might as well just change the define

morrowwolf (changes requested):
Otherwise looks good

code/modules/mob/living/carbon/xenomorph/ai/movement/facehugger.dm
xDanilcusx: actually it's somewhat redundant

code/modules/mob/living/carbon/xenomorph/ai/movement/facehugger.dm
xDanilcusx: aye

morrowwolf (approved):

PR #23, opened 2023-10-22
Lurker's scary AI

xDanilcusx:
### and yes, they can taunt if someone keeps walking nearby for too long


morrowwolf:
As some notes to self:

Remove single letter vars
See if moving the copied runner code into a xeno proc would be better
Also make sure the lurker/runner crossover stuff is up to date

code/modules/mob/living/carbon/xenomorph/abilities/lurker/lurker_abilities.dm
```diff
@@ -30,6 +30,10 @@
 			if (istype(lurker_invis))
 				lurker_invis.invisibility_off()
 
+		var/datum/xeno_ai_movement/linger/lurking/AI = xeno.ai_movement_handler
```
morrowwolf, line 33: Oh and note to self see if we can't move this somewhere where a signal may work, I don't like these movement calls like this as we may have switching behaviors eventually

xDanilcusx: huh alright

PR #33, opened 2023-10-29
Vehicle autofire

Doubleumc:
# About the pull request

Convert vehicle hardpoints from using their bespoke firing system to one structured closely on handheld guns and deployables such as the M2C. Now using the `autofire` component. Much like handheld weapons it is capable of different firemodes (semi/burst/auto) and changing targets during fire.

Hardpoints were converted to match their old effectiveness as closely as possible; this is intended as a quality of life improvement, not a rebalance. Damage, AP, range, ammo, etc were not touched.

Fire rates were copied over directly. Single-fire weapons with long delays were made semi-auto (e.g. LTB), and those with short delays were made full-auto (e.g. autocannon). Burst-fire weapons with significant extra delays after the burst remained burst-fire (cupola, smokescreen), and the rest were converted to full-auto (e.g. dual cannon). While changing firemodes is easily implemented, no weapon seemed a good candidate for more than one firemode and so that is omitted for now.

Scatter was approximated. The existing `accuracy` functioned as a percent chance the shot would stray one tile from the target. Gun-style `scatter` is instead a cone of fire in degrees. No direct conversion is possible, so scatter values are roughly set such that firing at a tile at the edge of the screen should "feel" about as accurate. Closer ranges would experience less spread than before, longer ranges more.

The buffing weapon sensor module was adjusted to work with the new firing system, and effects hardpoint scatter angle and firing rate. Vehicle buffs still use multipliers instead of adding/subtracting as handheld guns do, as a flat +/- adjustment to fire delay would have a significantly different effect on slow firing weapons (e.g. LTB) vs fast firing (e.g. autocannon). One major difference is that burstfire delays are effected and buffs increases the burst density. Before, there was a single cooldown initiated at the start of the burst, and only that cooldown was modified by the buff. Now, since the inter-burst delay is needed by the `autofire` component both the inter-burst delay and the after-burst delay are modified by buffs.

Activating non-selected hardpoints was removed as not compatible. The issue is that tracking a single click's modifiers is no longer sufficient, it has to track through the whole mousedown-to-mouseup period and the user can change multiple click modifiers in that time. I could not find a method that was satisfactory without a much bigger overhaul of vehicle controls than I'd like to take on in a PR not meant for it. I'm sure it can be done, but that brings up the question of if that's even the control scheme we'd want, in a PR that was never meant to ask that question let alone answer it.

Vehicle weapons using `gun`-like code makes them easier and more familiar to use, and more code commonality makes maintenance just a little bit easier.

<details>

</details>

# Changelog

:cl:
refactor: vehicle weapons can fire full-auto
del: no more controls for firing vehicle non-selected weapons
/:cl:


code/modules/vehicles/multitile/multitile_interaction.dm
```diff
@@ -337,6 +337,90 @@
...
+				//activate_support_module(source, seat, object)
+				return
+		if(VEHICLE_SUPPORT_GUNNER_ONE, VEHICLE_SUPPORT_GUNNER_TWO)
+			if(modifiers["shift"])
+				//object.examine(source)
+				return
+			if(modifiers["middle"] || modifiers["alt"] || modifiers["ctrl"])
+				return
+
+	if(modifiers["shift"] || modifiers["middle"] || modifiers["right"])
```
fira, line 413: You should use the modifier defines instead of strings, it reduces chance of mistakes

fira (commented):
Pretty cool. Could I convince you to eventually PR it upstream?

code/modules/vehicles/multitile/multitile_interaction.dm
Doubleumc: Didn't know those were a thing. Swapped over now.

Doubleumc:
I do intend to PR it upstream when its ready. Not quite at feature parity yet, need to wrangle the hardpoint buffs and the controls for not-selected hardpoints.

code/modules/vehicles/hardpoints/hardpoint.dm
```diff
@@ -166,37 +194,61 @@
...
 	if(LAZYLEN(type_multipliers))
 		for(var/type in type_multipliers)
-			V.dmg_multipliers[type] *= LAZYACCESS(type_multipliers, type)
+			vehicle.dmg_multipliers[type] *= LAZYACCESS(type_multipliers, type)
 	if(LAZYLEN(buff_multipliers))
 		for(var/type in buff_multipliers)
-			V.misc_multipliers[type] *= LAZYACCESS(buff_multipliers, type)
+			vehicle.misc_multipliers[type] *= LAZYACCESS(buff_multipliers, type)
 	buff_applied = TRUE
+	vehicle.on_modifiers_change() //check if I need to do this? currently only "accuracy" and "cooldown" need it
```
fira, line 220: did you check? if you wanna leave it like this could rewrite comment to be less ambiguous

code/modules/vehicles/hardpoints/hardpoint.dm
```diff
@@ -166,37 +194,61 @@
...
 	if(LAZYLEN(type_multipliers))
 		for(var/type in type_multipliers)
-			V.dmg_multipliers[type] *= 1 / LAZYACCESS(type_multipliers, type)
+			vehicle.dmg_multipliers[type] *= 1 / LAZYACCESS(type_multipliers, type)
 	if(LAZYLEN(buff_multipliers))
 		for(var/type in buff_multipliers)
-			V.misc_multipliers[type] *= 1 / LAZYACCESS(buff_multipliers, type)
+			vehicle.misc_multipliers[type] *= 1 / LAZYACCESS(buff_multipliers, type)
 	buff_applied = FALSE
+	vehicle.on_modifiers_change() //check if I need to do this? currently only "accuracy" and "cooldown" need it
```
fira, line 233: same here

code/modules/vehicles/hardpoints/hardpoint.dm
```diff
@@ -490,76 +500,206 @@
...
+/obj/item/hardpoint/proc/get_origin_turf()
+	return get_offset_target_turf(get_turf(src), origins[1], origins[2])
+
+/// Plays 'click' noise and announced to chat. Usually called when weapon empty.
+/obj/item/hardpoint/proc/click_empty(mob/user)
+	if(user)
+		to_chat(user, SPAN_WARNING("<b>*click*</b>"))
+		playsound(user, 'sound/weapons/gun_empty.ogg', 25, 1, 5) //5 tile range
+	else
+		playsound(src, 'sound/weapons/gun_empty.ogg', 25, 1, 5)
```
fira, line 677: the hardpoint is outside the vehicle, and the user inside, right? why not both?

code/modules/vehicles/hardpoints/hardpoint.dm
```diff
@@ -91,24 +79,63 @@
...
+	var/burst_delay = 1
+	/// When burst-firing, this number is extra time before the weapon can fire again.
+	var/extra_delay = 0
+	/// If the gun is currently burst firing.
+	var/burst_firing = FALSE
+
+	/// Currently selected target to fire at. Set with set_target().
+	var/atom/target
+	/// Current operator (crew) of the hardpoint.
+	var/mob/hp_operator
```
fira, line 129: unless i'm missing something ideally you'd hook COMSIG_PARENT_QDELETING on the operator and reset_fire if they get deleted to avoid loose references. i know it's a pretty edge case but when it comes to mob references every bit counts

fira (commented):
Looks good to me, some notes I can make

code/modules/vehicles/hardpoints/hardpoint.dm
Doubleumc: For clarity `on_modifiers_change()` should be called whenever a buff modifier is changed, so it can inform all the hardpoints to update their fire rates/etc. What I meant to say (to myself) with that comment was "should this check if a value was actually changed before potentially calling a proc on every hardpoint?" Considering how rarely buffs change and the most hardpoints any vehicle has is 5 its probably not a concern. I'll remove the comment.

code/modules/vehicles/hardpoints/hardpoint.dm
Doubleumc: I didn't think about this, was just directly copied from gun code. Considering internal sounds can be heard externally and vice-versa the difference is minimal, but for consistency's sake with the other hardpoint sounds I'll have the sound played externally.

code/modules/vehicles/hardpoints/hardpoint.dm
Doubleumc: That case is currently handled indirectly. `hp_operator` is set by the hardpoint when firing starts, and cleared when firing ends. If an operator were deleted that would cause them to be unbuckled, eventually calling `on_unset_interaction()`, aborting fire, and the hardpoint would clear `hp_operator`. [Here:](https://github.com/PvE-CMSS13/PvE-CMSS13/pull/33/files/a418e1bb33744df8132ddbfc691d0d9992b0dab1#diff-8e099084a34f9233ac761a049a49511a3239b14470147fe2cad7dd464d3de9b0R345-R350)
image

Doubleumc: On further thought, since `multitile` already does the gruntwork for tracking who's in what seat its unnecessary complexity for `hardpoint` to independently track its user. Removed `hp_operator` and instead pulls from `multitile`'s `seats[]` list of crew when necessary.

PR #37, opened 2023-11-01
Midway

morrowwolf:
I can't tell if this is a new github style or if my browser is bugging out. Does that mean it's a good style? Change is hard honestly.

Anyways this is the new and smaller dropship for the new player size. It's by no means final and anyone is free to take a crack at making it better. 

Specifically the front looks a bit... fat? I'm not sure.

Oh and we've lost our electronic slots but I'm not sure we need them in our use case.

There will likely be some bugs and such as there are fifty thousand systems all interacting with dropships.

Oh and the cameras on the Golden Arrow are set right now.

Shout out to Rocknshoulder on the discord for doing 90% of the work.


maps/shuttles/dropship_midway.dmm
```diff
@@ -0,0 +1,815 @@
...
+	},
+/area/shuttle/midway)
+"dr" = (
+/obj/structure/machinery/door_control{
+	id = "dropship_midway";
+	name = "Dropship Lockdown";
+	normaldoorcontrol = 3;
+	pixel_y = -19;
+	req_one_access_txt = "3;22";
+	throw_range = 15
```
fira, line 23: throw_range??

morrowwolf: WHAT DA HECK

PR #40, opened 2023-11-02
Slotting Rework part 1

morrowwolf:
This PR changes how jobs are assigned and cycled.

Instead of cycling:
Priorities
Jobs
Players

We now cycle:
Players
Priorities
Jobs

This should stop some of the 4d chess considerations of trying to figure out who is rolling for what and trying to high roll for a slot that has the least people rolling for it.

Now, you just have a (slots available / players readied) chance to get in if you have every job at least low priority.

This will need a good bit of testing (as all slotting changes generally do). As a note, the debug logs will be removed after testing is complete.

More changes to job selection likely coming.


morrowwolf:
OK I've decided something about the debug logs in STUI makes neurons fire in my brain in a nice way so I'm keeping them

PR #43, opened 2023-11-03
Possible sentry message lag fixes

morrowwolf:
Possible sentry message lag fixes.

May be part of the issue or perhaps I made it worse we'll see


morrowwolf:
Seems to help surprisingly 🤷

PR #44, opened 2023-11-03
Xeno pathing tweak that may destroy us all

morrowwolf:
This is either going to make the pathing far better for chokes or it's going to make the xenos act in some extremely goofy ways

We'll find out


morrowwolf:
Seems to have made it waaaay better

PR #49, opened 2023-11-05
Adds pilot, crew chief, synth to low pop

WolfOrion:
Adds pilot, dropship crew chief, and synth back to role selection. Also adds more seats to the Midway, as well as a dropship parts printer and a pilot and crew chief bunkroom.


WolfOrion:
THIS IS NOT MY FORK, SORRY

PR #54, opened 2023-11-06
Lurkers hiding on structures and minor xeno interactions tweaks

xDanilcusx:
## yeeeeah maybe it's still too early for structure lurking but im too up for it :p


code/modules/mob/living/carbon/xenomorph/ai/movement/lurking.dm
```diff
@@ -67,7 +73,12 @@
...
 
 		var/blocked = FALSE
-		for(var/atom/potential_blocker as anything in potential_home)
+		for(var/obj/structure/potential_blocker in potential_home)
+			if(potential_blocker.unslashable && (potential_blocker.can_block_movement || potential_blocker.density))
+				blocked = TRUE
+				break
+
+		for(var/mob/potential_blocker in potential_home)
 			if(potential_blocker != idle_xeno && (potential_blocker.can_block_movement || potential_blocker.density))
```
fira, line 82: i know it's the same throughout the whole file, but movement code works off `can_block_movement && density` as a condition rather than || - Items for example can be dense but not block movement and be passed over, and same for humans that when falling down are still dense but `can_block_movement` but you can still walk over them

morrowwolf:
Target stuff will be killed from here: https://github.com/PvE-CMSS13/PvE-CMSS13/pull/51/commits/2e062cbd02866dd4e8539cddcd9742bacafc6e6e

code/modules/mob/living/carbon/xenomorph/ai/movement/lurking.dm
xDanilcusx: #1442fe3

PR #59, opened 2023-11-08
Surface icon baking removal

morrowwolf:
Removes surface icon baking

Now I don't have to VV the table to move to the turf to click on shit

I dunno if this will effect performance or potentially blow up the universe so we'll test first


fira:
It was supposed to be a performance change to reduce load in SendMaps - now that we have it multithreaded it's probably not as much of an issue. Let us know!

PR #70, opened 2023-11-20
Calculate path delay for no path found situations

morrowwolf:
Adds a delay to calculating a new path when we could not find a path.

This should minimize new game masters (or me doing stupid stuff) blowing up xeno pathfinding by accident.

Could be upped in the future if it continues to pop up as an issue.


morrowwolf:
bro

PR #74, opened 2023-11-23
GM button: delete nearby xenos

Doubleumc:
# About the pull request

When you *don't* want to kill every motherfucker alive. Deletes xenos that are in the GM's viewport.
image

Adds an intermediate option between deleting all xenos and having to right-click>Delete every bug in a swarm.

<details>

</details>

# Changelog

:cl:
add: GM option to delete nearby xenos
/:cl:


morrowwolf (changes requested):
Reposting from discord:

A few things:

Avoid usr like the plague. It’s inconsistent. In this case use ui.user instead as it will always be good to go (hypothetically).

Second, for anything longer than like a line or two (and really they should be as well) put it in a proc that this calls so we separate the logic from the act handler in case we want to reuse it elsewhere.

Third, I would use world.view as your reference distance and then cycle through all xenos and do a get_dist(xeno, user) and if less than world.view delete.

OR

You can do:
for(var/mob/living/carbon/xenomorph/cycled_xeno in ui.user.view())

I honestly don’t know which way would be more performant. Probably depends on how many xenos are spawned but I usually trust get_dist() and cycling through the global list as get_dist() is pretty damn fast afaik

Fourth, split up the early return checks into their own lines just for readability sake.

Doubleumc:
Switched to using `view()` and modeled it after the "InView" section of the Admin tab. If its performant enough for admins of a 200-some player server, its good enough for us.

Funnily enough, in testing I discovered the existing "InView" actions use the default view, not the user's current view. This one will correctly use the user's current view.

morrowwolf:
> If its performant enough for admins of a 200-some player server, its good enough for us.

FWIW lots of code in the codebase is trash but this proc is incredibly cold so no reason to really worry too much about the details I suppose.

morrowwolf (approved):

PR #75, opened 2023-11-23
GM button: toggle vehicle blockers

Doubleumc:
# About the pull request

A convenient button to enable/disable those invisible vehicle blockers.

image
image

Significantly more convenient than the existing methods.

<details>

</details>

# Changelog

:cl:
add: gm button to toggle vehicle blockers
/:cl:


morrowwolf (changes requested):
Instead of having it be a mode toggle I'd far prefer a global list of all vehicle blockers that have an active var that we toggle on/off.

We're confined to a certain number of bitflags and we're starting to walk closer to the limit.

Secondly, and I'm happy to do this if you don't get to it, could you add another section that's just "flags" or "toggles". Should be just changing the categories of this and the other one to "Game Master.Toggles"

Doubleumc:
Changed it to work like "Toggle LZ Weeding"/"Toggle OB Spawn" where it toggles a global variable that the effected system checks.

Moved "Toggle Vehicle Blockers" and "Toggle Xeno Spawning" to the Flags section to mirror the Admin tab.

code/modules/admin/admin_verbs.dm
```diff
@@ -329,6 +329,7 @@ var/list/roundstart_mod_verbs = list(
 		add_verb(src, /client/proc/toggle_game_master)
 		add_verb(src, /client/proc/toggle_join_xeno)
 		add_verb(src, /client/proc/game_master_rename_platoon)
+		add_verb(src, /client/proc/toggle_vehicle_blockers)
```
morrowwolf, line 332: Also needs removed line 364 for de-admin'ing and such

morrowwolf (dismissed):
Otherwise fine

code/modules/admin/admin_verbs.dm
Doubleumc: Done.

PR #96, opened 2024-01-24
pain code minor cleanup

private-tristan:
# About the pull request
removes some useless signal removals because the signals themselves weren't registered

while I didn't record it, I did boot up a server and made sure that devouring still killls with 200 pain, and that when i set pain to 0 it doesn't.
</details>

# Changelog

no player facing changes.


FslashN (approved):
Looks fine. There is only signal handler, so it does seem like the other stuff was either left over from something or didn't get implemented in the first place.

PR #111, opened 2024-02-11
adds crappy system for delayed/looping spawns

tmtmtl30:
# About the pull request

Adds buttons to the GM panel allowing for xenomorph spawns to be delayed by up to 1 hour, "looped" so that they repeat on the same delay, and hidden from players by failing if any living, connected humans are found within 9 tiles (this effect applies for non-delayed spawns as well, if the toggle is pressed). These spawns can then be teleported to or deleted remotely via the panel.

the approach here is sort of crappy, but i'm not sure whether /obj/effect s are as ubiquitous here for effects like these as they are on other codebases, or how to make effects visible to GMs only. 

Makes it easier for GMs to juggle multiple things at once by letting them set up spawns in advance, while preventing those spawns from breaking the fourth wall.

<details>

i tested it well enough. didnt take screenshots and i am currently very tired

</details>

# Changelog

:cl:
add: Added delayed, looped, and "hidden" xenomorph spawns for GMs.
/:cl:

### DIFF
--- a/code/modules/admin/game_master/game_master_xeno_spawner.dm
+++ b/code/modules/admin/game_master/game_master_xeno_spawner.dm
@@ -1,0 +1,103 @@
+GLOBAL_LIST_EMPTY(gm_xeno_spawners)
+
+/datum/game_master_xeno_spawner
+	/// The actual image holder that is added to game masters' clients
+	var/image/gm_image
+
+	/// The location the xenomorphs will spawn at.
+	var/turf/spawn_loc
+
+	/// The type of xeno to be spawned.
+	var/mob/living/carbon/xenomorph/spawn_type
+
+	/// The number of xenos to be spawned.
+	var/spawn_count
+
+	/// The delay until xenos are spawned (may be 0, in which case xenos spawn instantaneously).
+	var/spawn_delay
+
+	/// If TRUE, the spawner will not delete itself after firing.
+	var/looping
+
+	/// The hive the spawned xenos will belong to.
+	var/spawn_hive
+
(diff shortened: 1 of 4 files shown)


code/modules/admin/game_master/game_master_xeno_spawner.dm
```diff
@@ -0,0 +1,103 @@
...
+	// cleans up the image
+	if(gm_image)
+		for(var/client/game_master in GLOB.game_masters)
+			game_master.images -= gm_image
+		QDEL_NULL(gm_image)
+
+/datum/game_master_xeno_spawner/proc/spawn_xenos()
+	var/failure = FALSE
+
+	if(fail_if_human_near)
```
FslashN, line 90: If `fail_range` exists and is not a null, `fail_if_human_near` may not be needed as the range already signals that it should do something within that range. Could maybe change the range instead of toggling the function, though in practice you probably would rarely want it to work like that.

FslashN (commented):
This has been up on TR for a while, and we've got some feedback about the feature. It can be useful as-is, but the usefulness is limited as you still have to babysit the spawner to an extent, which takes away from the GM's attention.

- It would likely work better if it was tied to vents and tunnels, or had the option to be tied into them.
- Ability to middle click remove the spawner in addition to the menu.
- Ability to spawn only a # of times specified, after which the spawner auto-deletes.
- Ability for it to auto-delete when a marine moves into specified range.
- Ability to modify the range for marine detection.

FslashN:
This has been up for a while and doesn't seem like it's being actively developed. We can re-open this in the future if desired.

PR #117, opened 2024-02-17
MK1 Mag Boxes and 10x28mm ammo cans

grifp71:
For some reason, these items were both missing from base CM code. This allows empty ammo cans for 10x28mm (smartgun ammo) to be crafted from cardboard and an empty mag box for MK1 magazines which will be a massive quality of life increase as our ammo just sits on the dropship to be flamed. This should theoretically make life even the slightest bit easier for GMs when it comes to resupplying platoons in round as well. MK1 boxes have been requested a few times from the community.

### DIFF
--- a/code/game/objects/items/stacks/sheets/sheet_types.dm
+++ b/code/game/objects/items/stacks/sheets/sheet_types.dm
@@ -280,6 +280,8 @@ var/global/list/datum/stack_recipe/cardboard_recipes = list ( \
 		new/datum/stack_recipe("empty rifle ammo box (10x24mm Incen)", /obj/item/ammo_box/rounds/incen/empty), \
 		new/datum/stack_recipe("empty rifle ammo box (10x24mm LE)", /obj/item/ammo_box/rounds/le/empty), \
 		null, \
+		new/datum/stack_recipe("empty rifle ammo box (10x28mm)", /obj/item/ammo_box/rounds/smartgun/empty), \
+		null, \
 		new/datum/stack_recipe("empty box of MREs", /obj/item/ammo_box/magazine/misc/mre/empty), \
 		new/datum/stack_recipe("empty box of M94 Marking Flare Packs", /obj/item/ammo_box/magazine/misc/flares/empty), \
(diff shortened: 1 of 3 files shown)


code/game/objects/items/stacks/sheets/sheet_types.dm
```diff
@@ -279,6 +280,8 @@ var/global/list/datum/stack_recipe/cardboard_recipes = list ( \
 		new/datum/stack_recipe("empty rifle ammo box (10x24mm Incen)", /obj/item/ammo_box/rounds/incen/empty), \
 		new/datum/stack_recipe("empty rifle ammo box (10x24mm LE)", /obj/item/ammo_box/rounds/le/empty), \
 		null, \
+		new/datum/stack_recipe("empty rifle ammo box (10x28mm)", /obj/item/ammo_box/rounds/smartgun/empty), \
```
FslashN, line 283: Shouldn't this be drum or something like that?

code/modules/projectiles/ammo_boxes/round_boxes.dm
```diff
@@ -40,6 +40,19 @@
 /obj/item/ammo_box/rounds/heap/empty
 	empty = TRUE
 
+//----------------10x28mm Ammunition Boxes (for smartguns)------------------
+
+/obj/item/ammo_box/rounds/smartgun
+	name = "\improper rifle ammunition box (10x28mm)"
```
FslashN, line 46: And this one here?

FslashN (commented):
I think leaving it as rifle ammunition is a little confusing.

code/game/objects/items/stacks/sheets/sheet_types.dm
grifp71: > Shouldn't this be drum or something like that?

This particular one is a box of loose ammo, not drums, that you use to repack drums.

code/modules/projectiles/ammo_boxes/round_boxes.dm
grifp71: Made tweaks. I think I did this correctly? Not sure, still working my way through figuring out github properly.

PR #118, opened 2024-02-18
Update ice_colony_v2.json

FslashN:
Fixes the Ice v2 runtime where the map JSON failed to load due to a bad item path, thus being unable to actually select the map to play on.


FslashN:
Still need to figure out what's going on the the tests. There are no runtimes in-game, but there's something weird going on with the map item loading.

PR #121, opened 2024-02-20
Adds M48A4 to the character creation loadout

Merrgear:
# About the pull request

Pretty self explanatory, M48A4 is now in the character creation pointbuy

Muh two world wars! and i just think its neat. Allows a little more character to the marines choice of sidearms. +

# Changelog

:cl: Merrgear2
add: Adds the M48A4 Service Pistol to the loadout pointbuy
:cl:

### DIFF
--- a/code/game/objects/items/storage/misc.dm
+++ b/code/game/objects/items/storage/misc.dm
@@ -133,6 +133,24 @@
 	max_w_class = SIZE_MEDIUM
 	storage_slots = 3
 
+/obj/item/storage/box/M48A4_loadout
+	name = "M48A4 storage case"
+	desc = "A relatively large storage case containing the M48A4 and magazines. Purchased by enlisted or aspiring PMCs looking to carry a timeless classic"
+	icon = 'icons/obj/items/storage.dmi'
+	icon_state = "guncase"
+	w_class = SIZE_MEDIUM
+	max_w_class = SIZE_MEDIUM
+	storage_slots = 7
+
+/obj/item/storage/box/M48A4_loadout/fill_preset_inventory()
+	new /obj/item/weapon/gun/pistol/m1911/socom(src)
+	new /obj/item/ammo_magazine/pistol/m1911(src)
+	new /obj/item/ammo_magazine/pistol/m1911(src)
+	new /obj/item/ammo_magazine/pistol/m1911(src)
+	new /obj/item/ammo_magazine/pistol/m1911(src)
+	new /obj/item/ammo_magazine/pistol/m1911(src)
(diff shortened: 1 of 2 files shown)


code/game/objects/items/storage/misc.dm
```diff
@@ -133,6 +133,24 @@
...
+	name = "M48A4 storage case"
+	desc = "A relatively large storage case containing the M48A4 and magazines. Purchased by enlisted or aspiring PMCs looking to carry a timeless classic"
+	icon = 'icons/obj/items/storage.dmi'
+	icon_state = "guncase"
+	w_class = SIZE_MEDIUM
+	max_w_class = SIZE_MEDIUM
+	storage_slots = 7
+
+/obj/item/storage/box/M48A4_loadout/fill_preset_inventory()
+	new /obj/item/weapon/gun/pistol/m1911/socom(src)
```
FslashN, line 146: I don't like having the SOCOM variant available as a flavor gun; it's got several attachments and is meant to be better than the standard. I would make this the regular 1911, which is an iconic gun I can see people would want for flavor.

code/game/objects/items/storage/misc.dm
```diff
@@ -133,6 +133,24 @@
...
+	new /obj/item/ammo_magazine/pistol/m1911(src)
```
FslashN, line 147: Six mags is quite a few. I think two extra mags would be fine given that you can stuff the gun and the mags into a shoulder holster, just like the VP78.

code/game/objects/items/storage/misc.dm
Merrgear: the SOCOM 1911, as i recall has no stat changes from the normal 1911. the one in the PR isnt the outfitted one with the RDS suppressor and laser sight. and the main reason i chose the SOCOM one is the regular 1911 looks so fucking ugly. 
image
if there was an updated to not be shit sprite for the base 1911 i wouldve used that one, cause functionally they are the same

code/game/objects/items/storage/misc.dm
Merrgear: I just took the base kit you get from req blackmarket and put it into something that fits into the backpack that i think was meant to be used in a pistol belt. only around 42 rounds now that the 1911 mag gives 7 instead of 14 rounds, but i can change that

code/game/objects/items/storage/misc.dm
FslashN: The SOCOM variant is a little different stat wise; slightly more accurate with less scatter at the expense of a slightly lower damage multiplier.

You could add a "custom" variant that is just a reskinned 1911 with a new description, just like the other flavor guns.

PR #162, opened 2024-03-19
Special buckshot will now deal mostly normal buckshot 

KoishiVibe:
# Summary
- Special B/S will now not kill outright. Instead it will put people into various states of paincrit if they are wearing armor of any note, even at pointblank. 
- Multipliers ensure anti-xenomorph damage is retained
- Light armor will incap with fractures of the chest, medium is over 100 damage, heavy is somewhere around 50 damage per shot.
- Still needs adjusting based on vision for what is actually wanted, the idea of having to pay up for someone else's mistakes strikes me as a terrible game design decision and very unfun, but I do like the RP potential of "AUGH FUCK GODDAMNIT YOU SHOT ME IN THE BACK"


code/datums/ammo/bullet/shotgun.dm
```diff
@@ -154,14 +154,19 @@
...
 	max_range = 8
-	damage = 60
+	damage = 50
 	bonus_projectiles_amount = EXTRA_PROJECTILES_TIER_8
 	firing_freq_offset = SOUND_FREQ_LOW
+//special B/S type is super buffed vs xenomorphs but not as much against anything else
+/datum/ammo/bullet/shotgun/buckshot/special/on_hit_mob(mob/hit_mob, obj/projectile/hit_projectile)
+	if(isxeno(hit_mob))
+		var/mob/living/carbon/xenomorph/xeno = hit_mob
+		xeno.apply_damage(damage * 1.75, BRUTE)
```
FslashN, line 169: I'm really not a fan of variable damage based on mob type.

code/datums/ammo/bullet/shotgun.dm
```diff
@@ -190,13 +195,18 @@
...
 	accurate_range = 8
 	max_range = 8
-	damage = 90
+	damage = 25
 	firing_freq_offset = SOUND_FREQ_LOW
 
+/datum/ammo/bullet/shotgun/spread/special/on_hit_mob(mob/hit_mob, obj/projectile/hit_projectile)
+	if(isxeno(hit_mob))
+		var/mob/living/carbon/xenomorph/xeno = hit_mob
+		xeno.apply_damage(damage * 3.50, BRUTE)
```
FslashN, line 208: This make it very difficult to see how much damage is actually being dealt, as both the main projectile and the spread deal different damage, and that damage is being dealt differently for xenomorphs.

FslashN (changes requested):
Just from personal opinion, I don't like this change. If you're doing huge damage to aliens, you should be doing huge damage to allies if you misfire. Friendly fire is one reason marines can't just spam their guns in a lot of instances. Also, logically, there is no reason special buck would do more damage to aliens than marines, unless it was designed to specifically kill xenos.

Having said that, I do think being instantly killed by special buck is kind of unfun, since it can happen suddenly and with no warning to the person getting shot. On a technical side, variable damage numbers like this lead to confusing and inconsistent math (and snowflake code); some ways you could handle it:

Nerf buck so that it doesn't immediately kill marines, though I doubt this is feasible while keeping it at the same power level against aliens.

Rework alien defenses to where multiple pellet weapons are uniformly more powerful against them, not just special buck itself. Maybe something to do with alien armor calculations.

Add some kind of PB prevention for friendlies, since that seems to be the most prevalent way for marines to insta die to the special buck. Though I don't know if that's necessarily mechanically sound either.

code/datums/ammo/bullet/shotgun.dm
KoishiVibe: Alright then. 16 damage base, 9 pellets in all, and the damage is multiplied by 5 across everything.

code/datums/ammo/bullet/shotgun.dm
KoishiVibe: I know, but it's the only way to make this work without making the armor super ultra high or mashing the reduce HP button...

or making it have more projectiles.

DexterDude:
Shotgun is faaaarrr too weak against marines. I got PBed and had 0 damage to my name except for shrapnel.

code/datums/ammo/bullet/shotgun.dm
```diff
@@ -154,14 +154,19 @@
...
-	name = "buckshot shell, USCM special type"
+	name = "special type buckshot"
 	bonus_projectiles_type = /datum/ammo/bullet/shotgun/spread/special
 
 	accurate_range = 8
 	max_range = 8
-	damage = 60
-	bonus_projectiles_amount = EXTRA_PROJECTILES_TIER_8
+	damage = 25
+	bonus_projectiles_amount = EXTRA_PROJECTILES_TIER_18
```
morrowwolf, line 163: This will likely cause lag weirdness, proooobably not this high lmao

I like what you're going for here, the decreased damage per pellet is probably too low if we lower the projectiles down again.

morrowwolf (changes requested):

AndroBetel (commented):
i have to agree with rock on this one, and apart from "high risk - high reward" thing we have with buckshot its kind of iconic to pve community and, imo, its not often marines get one shot killed by it

AndroBetel:
author decided to take a break from ss13, if they return they can contact any maintainer to have their PR reopened

PR #164, opened 2024-03-20
AI tweaks and fixes

xDanilcusx:
# About the pull request

A few more tweaks and fixes for AI interactions and loops

<details>

</details>

# Changelog

:cl:
add: AI will try and avoid stepping in toxic water
code: Xeno resting now isn't instant in most cases
code: Sentry xeno_ai_obstacle() now returns SENTRY_PENALTY instead of VEHICLE_PENALTY
fix: check_special_blockers() now properly checks for types of types listed in XENO_AI_SPECIAL_BLOCKERS
fix: AI will now act less stupid when it comes to poddoors and shutters
fix: Xenos are no longer swipe on floor for no reason
refactor: Penalty defines now placed in increasing order
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/xenomorph/xeno_ai_interaction.dm
+++ b/code/modules/mob/living/carbon/xenomorph/xeno_ai_interaction.dm
@@ -122,6 +122,7 @@ At bare minimum, make sure the relevant checks from parent types gets copied in 
 
 	return
 
+
 /////////////////////////////
 //          MOBS           //
 /////////////////////////////
@@ -218,7 +219,7 @@ At bare minimum, make sure the relevant checks from parent types gets copied in 
 	if(!.)
 		return
 
-	return VEHICLE_PENALTY
+	return SENTRY_PENALTY
 
 
 /////////////////////////////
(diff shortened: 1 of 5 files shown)


code/modules/mob/living/carbon/xenomorph/xeno_ai_interaction.dm
```diff
@@ -272,10 +284,30 @@ At bare minimum, make sure the relevant checks from parent types gets copied in
...
 	return INFINITY
+
+
+/////////////////////////////
+//          RIVER          //
+/////////////////////////////
+/turf/open/gm/river/xeno_ai_obstacle(mob/living/carbon/xenomorph/X, direction, turf/target)
+	. = ..()
+	if(!. && !covered)
+		. *= base_river_slowdown
```
morrowwolf, line 307: Should this not be +=?

If !. is true to get here then multiplying it should do nothing right?

morrowwolf (changes requested):

code/modules/mob/living/carbon/xenomorph/xeno_ai_interaction.dm
xDanilcusx: yep, that one wasn't intended to be with a NOT operator. Also switching to += as it makes more sense when we remember that our only parent proc returns either 1 or 0

morrowwolf (approved):

PR #177, opened 2024-03-24
Drakemaxxing or weapons on back for smartgunners

grifp71:
https://youtu.be/hXv3fqUwCcg?t=30

Allows smartgunners to utilize the backslot PURELY for weapons by proccing for weapon/gun.

This has been suggested for _months_ because you see Drake switch to a M240 in the movie. I have mained smartgunner pretty frequently on PvE (I nearly have my gold medal). I was initially opposed to the idea, but people keep asking for it - and people keep saying they're going to work on it, but nobody has and it keeps coming up, so I figured I'd take my first deep dive into coding beyond super simple edits and do it.

Now, I was initially opposed to this idea because I don't want to see smartgunners overshadow the other classes, but I'm not so sure it will be that big of a deal. If you take extra ammo for anything that isn't a SG, you're gimping your storage horribly. I like this for emergency situations and when everything is going to hell - just like Drake in the movie. The smartgun honestly starts to feel a bit anemic compared to the M41, so this adds a slight bit more pizazz to the role. (the volume of fire is nice, but holy hell does the M41 drop a xenoid faster)

Potential cons are smartgunners hogging the M37 and M240, but any squad sergeant or gunny worth their salt should smoke them if that happens, and honestly.. a lot of the time our riflemen fail to even carry the special weapons in the first place, so I guess if they won't take them, it's better to have them on the back of a smartgunner for situational use.

Now, I'm not smug enough to say that I think this should be merged for certain. As the author of this PR, I heavily advocate a test merge and see how it goes in actual gameplay. If it becomes a monster or people are stupid with it, we can can the idea and I'll be the first to agree. But I think it's movielike and has some soul, and our population is probably mature enough to handle it without hogging shit.

Tested locally, works, and of course smartgunners can still not wear satchels or backpacks.

Real credit goes to Andro for helping guide me through my first code dive. Thanks.

### DIFF
--- a/code/modules/clothing/suits/marine_armor.dm
+++ b/code/modules/clothing/suits/marine_armor.dm
@@ -424,6 +424,10 @@
 	if(slot != WEAR_BACK)
 		return
 
+
+	if(istype(equipping_item, /obj/item/weapon/gun))
+		return
+		
 	. = COMPONENT_HUMAN_CANCEL_ATTEMPT_EQUIP
 
 	if(equipping_item.flags_equip_slot == SLOT_BACK)


KoishiVibe:
The smartgun is supposed to holster to the back anyways. What about holstering the smartgun to the back and letting the suit slot take some small pouches such as for pistol ammo, or a pistol holster, or just the pistol itself?

grifp71:
> The smartgun is supposed to holster to the back anyways. What about holstering the smartgun to the back and letting the suit slot take some small pouches such as for pistol ammo, or a pistol holster, or just the pistol itself?

don't want to give them more storage and morrow is also very anti-storage

KoishiVibe:
> > The smartgun is supposed to holster to the back anyways. What about holstering the smartgun to the back and letting the suit slot take some small pouches such as for pistol ammo, or a pistol holster, or just the pistol itself?
> 
> don't want to give them more storage and morrow is also very anti-storage

In that case, then just a pistol or the sidearm pouch then?

code/modules/clothing/suits/marine_armor.dm
```diff
@@ -424,6 +424,10 @@
 	if(slot != WEAR_BACK)
 		return
 
+
+	if(istype(equipping_item, /obj/item/weapon/gun))
```
morrowwolf, line 428: This should be a list of allowed types that it cycles through on the smartgun rather than directly coded in for future maintainability.

morrowwolf (changes requested):

morrowwolf:
Not 100% sure about this one. I uuuh... eeeh? I'll think on it.

KoishiVibe:
if this doesn't go through, could I borrow some of the code?

morrowwolf:
Waiting on requested changes still.

code/modules/clothing/suits/marine_armor.dm
```diff
@@ -396,6 +396,13 @@
 		/obj/item/device/motiondetector,
 		/obj/item/device/walkman,
 	)
+	var/list/smartgun_back = list(
+	/obj/item/weapon/gun/shotgun/pump/special,
```
morrowwolf, line 400: Tab in one more

morrowwolf (changes requested):

morrowwolf (approved):

PR #178, opened 2024-03-25
Movie APC Mk2

DexterDude:
Updates the Movie APC (M577A3) to look much better and feel more like an APC.

Changes include:
- Smaller interior, much more cramped than the old movie APC, now the widest you have is a 2 wide hall down the middle of it to the driver seat and loader.
- Updated Platoon Commander area, new unique sprites for the Tactical Operations Center
- New overwatch console for the TOC
- New table for the TOC
- Dropship and tactical map computers are now integrated into the TOC overwatch console sprite as screens, though they still function the same as before
- New sprites for lockers, kinda make more sense than the weird squares.
- Lockers now are no longer job restricted, anyone can access them
- Lockers now also have the ability to store ammo boxes, no more ammo box clutter everywhere
- Pulse rifle rack for spare weapons, replacing the old vendor
- Ammo racks for the APC's weapons, unique sprites and all.

Also fixed the lack of the classic metallic "crunch" noise that the standard dropship seats still have, felt somewhat relevant to the PR.

I've also given the Platoon Commander the ability to drive the APC (skill 2), as previously they weren't able to due to their skill of 1 overriding the default skill of 2.

movieapcinterior


DexterDude:
Forgot to mention in the original post, but this has an issue where if the Platoon Commander seat is occupied then it's unable to be passed on the tile to the left of it since it technically has a person in it. I'd greatly appreciate it be fixed since the space here is very cramped. Just having it so that no matter if the seat is occupied or not it will always not be dense. Would appreciate if someone could somehow fix this for me.

Also, the TOC console is 2 wide but only able to be accessed by the tiled adjacent to the left side of it, would be great if that got fixed too, unsure how to do so.

FslashN:
This looks great. You've really polished up the older concept.

AndroBetel (changes requested):

AndroBetel (approved):

PR #185, opened 2024-03-27
Adds boiler suits to Platoon Commander lockers

DexterDude:
For platoon commanders who wish to deploy and want to LARP as gorman. The service uniform is too formal and the standard uniforms don't quite look too good without any armor on. Boiler suit is a nice middle ground. Also its cool.


FslashN:
I think this is fine.

PR #187, opened 2024-03-27
SA .357 Swap

Max-023:
Adds the new SA .357 autorevolver to Marine armories, replacing the M44 for a lore friendly option. Has an altered CMB sprite based on its Dark Descent variation, and comes with regular ammo rather than the hollow point that the CMB gets. Also ups the damage on the regular ammo to be in line with the M44, to avoid an ill felt balance difference.


KoishiVibe:
hm yes comically powerful .357

code/datums/ammo/bullet/revolver.dm
```diff
@@ -120,15 +120,15 @@
 	name = "small revolver bullet"
 	headshot_state = HEADSHOT_OVERLAY_LIGHT
 
-	damage = 45
+	damage = 70
```
FslashN, line 123: I'm not sold on this change. It can do more damage than a standard pistol, but 70 is pushing it with our other damage numbers. Maybe 55 or so might be fine.

code/datums/ammo/bullet/revolver.dm
```diff
@@ -120,15 +120,15 @@
...
 
 	penetration = ARMOR_PENETRATION_TIER_3
 
 /datum/ammo/bullet/revolver/small/hollowpoint
 	name = "small hollowpoint revolver bullet"
 	headshot_state = HEADSHOT_OVERLAY_MEDIUM
 
-	damage = 75 // way too strong because it's hard to make a good balance between HP and normal with this system, but the damage falloff is really strong
+	damage = 85 // way too strong because it's hard to make a good balance between HP and normal with this system, but the damage falloff is really strong
```
FslashN, line 131: Not really sure how to justify this. 85 damage for a hollowpoint seems very excessive. That's like sniper rifle tier damage.

FslashN (dismissed):

code/datums/ammo/bullet/revolver.dm
Max-023: The M44 does that damage already. Is that now a problem?

code/datums/ammo/bullet/revolver.dm
Max-023: It has the falloff modifier to make it drop damage significantly more, if the comments are still correct.

code/datums/ammo/bullet/revolver.dm
FslashN: M44 does 70?

FslashN: I looked it up and it sure does. I think that's nonsensical but it's not relevant to your pr.

Max-023: Yeah. It fires /datum/ammo/bullet/revolver which is the default template and has base damage of 70.

code/modules/projectiles/guns/revolvers.dm
```diff
@@ -746,3 +746,14 @@
 
 /obj/item/weapon/gun/revolver/cmb/normalpoint
 	current_mag = /obj/item/ammo_magazine/internal/revolver/cmb
+
+//-------------------------------------------------------
+//MARINE REVOLVER //Spearhead exists in Alien cannon, this one is based on the Dark Descent model.
+
+/obj/item/weapon/gun/revolver/cmb/marine
+	name = "\improper SA .357"
```
XSlayer300, line 754:
```suggestion
	name = "\improper .357 SA autorevolver"
```

code/datums/ammo/bullet/revolver.dm
KoishiVibe: the main thing is that FMJ has AP0 so you're really getting a weird effect

the other problem is the issue of actual ballistic effects on target. a pistol is just going to be worse in almost every way irl

to make it able to kill an xenomorph worth piss...

It Is What It Is ig.

code/modules/projectiles/guns/revolvers.dm
Max-023: SA stands for "Spearhead Autorevolver" so I don't think putting a second Autorevolver in the title is necessary.

XSlayer300: sometimes I really though "SA" would stand for "Spearhead Armories", hence the addition of the "autorevolver" on its title.

even the nsg23 commits this mistake by appending "assault rifle" after "NSG"
image

DexterDude:
was told to comment here, i dont really see this as a needed change, it feels like a change for the sake of change rather than something that actually impacts anything. i cant exactly say im a fan of the sprite either, it doesnt look as nice and refined as the current .44 sprite.

Merrgear:
codersprite in standard issue gives me pain

code/modules/projectiles/guns/revolvers.dm
Max-023: Fair enough. I'm just basing it on Dark Descent, which also calls the revolver "SA .357 Revolver". I can take on Revolver at the end.

Max-023:
As said, it's to replace the made-up M44 revolver from CM with a revolver that is actually from the setting and shown to be used by the Marines. The sprite is just a color swapped version of the CMB already in the game (with a slight silhouette change at the front to better match the pistol model). Considering other weapons have been swapped out by Morrow himself to keep in tune with the setting, such as the M41A Mk1 replacing the Mk2, I wouldn't say it is at all fair to say that adding the Marine service SA .357 is "change for the sake of change".

Just like the previous mention of the damage provided to normalpoint ammo, also I do not see how personal preference for sprite quality would bar a PR when the weapon its based on is already in the game and very present on the maps. There is a proposed new sprite in the CM discord, but I have no way of contacting the guy who made it for permission to use here.

code/modules/projectiles/guns/revolvers.dm
KoishiVibe: Wait it ISN'T Single Action .357? Topkek.

PR #188, opened 2024-03-27
more fluff, removes kevlar mentions and replaces with venlar

KoishiVibe:
# Summary
- most lore is reworked and a few really old mentions of jumpsuit are removed.
- kevlar replaced with venlar
- AVPwiki is wrong about the M10, we never actually find out how its made from the TMan.
- a couple of frankly fluffbased statistical changes


code/modules/clothing/under/marine_uniform.dm
```diff
@@ -226,7 +226,7 @@
 
 /obj/item/clothing/under/marine/officer/boiler
 	name = "marine operations uniform"
-	desc = "An operations uniform worn by members of the USCM. Do the corps proud. It has shards of light Kevlar to help protect against stabbing weapons and bullets."
+	desc = "Officer's uniform, a little outdated. Why it needs to be frag rated is beyond everyone."
```
FslashN, line 229: I'm not sold on stuff like this. The first part is informative, the second part is half personal opinion. Being frag rated is probably less to do with anything logical but some gameplay consideration. I'd prefer something along the lines of: "Officer's uniform, a little outdated, but fits surprisingly well."

code/modules/clothing/under/marine_uniform.dm
```diff
@@ -235,7 +235,7 @@
 
 /obj/item/clothing/under/marine/officer/command
 	name = "\improper USCM officer uniform"
-	desc = "The well-ironed utility uniform of a USCM officer. Even looking at it the wrong way could result in being court-martialed. It has shards of light Kevlar to help protect against stabbing weapons and bullets."
+	desc = "A commanding officer's uniform. The privileges of rank extend to having your uniform custom fitted apparently."
```
FslashN, line 238: And like this here. I think the personal opinion isn't really necessary. "A commanding officer's uniform. It is a custom fit, tailored to specification with great attention to detail."

FslashN (changes requested):

code/modules/clothing/under/marine_uniform.dm
KoishiVibe: I do actually have a question
what is the boiler uniform for

code/modules/clothing/under/marine_uniform.dm
KoishiVibe: thought it was a little fun tongue in cheek. I'll change it though.

FslashN: Tongue-in-cheek is fine if the item itself is supposed to be a little silly, or otherwise the tone shift is justified somehow. If you're describing something to the player, it should be mostly facts unless there is a specific reason to break away from that to make a joke.

code/modules/clothing/under/marine_uniform.dm
FslashN: Deploying into combat, I would imagine. But I can't say for sure.

FslashN:
I'll look into this again soon.

morrowwolf (approved):

PR #199, opened 2024-03-31
Jungle Update fix

AmoryBlaine:
Fixes minor icon error with Jungle update as well as incorrect boiler suit icon state.

# About the pull request

Fixes minor issues with jungle camo set from previous PR, regarding suit slot belt, as well as fixes sprite error I made a year ago with the boiler suit.


FslashN:
Seems like a minor fix.

PR #207, opened 2024-04-02
updated secure door icon palette

AmoryBlaine:
Updates the secure door palette to match that of the USCM ship palette.


FslashN:
This is actually good. The double doors and the single door versions were of different colors palettes, which made it confusing.

PR #212, opened 2024-04-04
Modifies some cosmetic effects, removes a little weirdness.

KoishiVibe:
# summary
- previously, if your damage after armor was <= 3 it would negate. this adds a little weirdness to some already slightly RNG affected damage. 
- now it only takes effect after being below 1, which might help with fractional damage weirdness.
- increases the thresholds at which armor sparks occur. it looks cool, but doesn't actually do anything. Immersion without having to modify all the armor values?


AndroBetel:
author decided to take a break from ss13, if they return they can contact any maintainer to have their PR reopened

PR #213, opened 2024-04-04
Biblically Accurate Smartguns

grifp71:
https://docs.google.com/document/d/19WB_cbJ1Yc-Z2byVt4XO66vzVr6A46fC54E8YYfivAs/edit?usp=sharing

Based off of a discussion today in the dev channel that included multiple people including myself, Andro, Morrow and Rock:

Significantly buffs Smartgun damage to be more lore accurate. This thing is a hammer. Fixes some of the fall-off. She's a monster, but you have to be supported by your team or you eat shit.

Shifts the smartgun into more of a crew-served weapon. It now requires an ammo bearer. The smartgunner cannot carry any extra drums. The squad has to do it for him in their satchels or backpacks. Soulful and makes sense as the squad is literally built around the smartgun.

Disallows the smartgun drum from being placed in GP pouches, the G8 belt, and drop pouches. Probably missed something but we'll patch any issues that arise.

Removes SG belts and adds 2 drums that spawned in the belts before hand.

This would work well hand-in-hand with #177 

These damage values are not final and need playtested, but I think this is a soulful, cool change that fits the lore and atmosphere and makes the smartgun more interesting.


code/datums/ammo/bullet/special_ammo.dm
```diff
@@ -11,17 +11,19 @@
...
 
 	accurate_range = 12
 	accuracy = HIT_ACCURACY_TIER_2
-	damage = 24
+	damage = 55
 	penetration = ARMOR_PENETRATION_TIER_8
+	damage_falloff = DAMAGE_FALLOFF_TIER_7
 	damage_armor_punch = 1
 
 /datum/ammo/bullet/smartgun/dirty
```
private-tristan, line 29: should also buff the other smartgun types?

grifp71: after testing, sure. letting this get playtested first and what not. presets also need worked on, that'll all come after testing

bearrrrrrrr:
uhhh, that's a huge buff
killing players without acid damage is already a huge problem, and crushers are already kind of on the weaker side now as is
not sure if this is a good idea

grifp71:
Damage adjusted since first iteration, getting positive feedback on the new values. Awaiting more testing.

bearrrrrrrr:
eghhh still not a fan but 42 is better than 50 i guess

now up to player preference so idm

Ollanius:
I've played a couple full-length rounds while this pull request was active.
Here's my personal thoughts on the matter:

- Increase the _'highly precise'_ bullet damage to **48-56**, and keep the _'armor shredding'_ bullet damage at **30**.
I know that it does a lot of damage, but I believe that _should_ be the case. It's a Smartgun, after all: a handheld autocannon that fires a larger cartridge than the standard Pulse Rifle. It's _the_ big fucking gun, for when you absolutely need to shred an oncoming stampede of unarmored bugs. If you want to easily counter a Smartgunner, send a Crusher their way. Those fuckers'll tank a hundred armor-shredding bullets without batting an eye, and will **always** end up killing _(or incapacitating)_ the Smartgunner before dying.

Following that train-of-thought, we should probably reduce the total amount of drums that the Smartgunner's locker spawns with **if** we raise the _'highly precise'_ bullet damage. Instead of six spare drums, we could drop it down to four spare drums?

- Allow for _'smartgun drums'_ to only be stored in _'drop pouches'_, _'medium general pouches'_, and the jumpsuit belt slot.
Smartgunners can comfortably carry about two spare drums with them, at most. You can theoretically carry up to five drums if you sacrifice your belt slot and pouches, but that means sacrificing your ability to carry additional utilities _(like a machete scabbard or pistol belt)_ and supplies _(like medicine and flares)_ With that in mind, most Smartgunners will burn through their personal supply within twenty minutes: this means that if a Smartgunner wants to actually stay effective for a whole mission, they'll **need** to coordinate with their squad.

You can see this with the _'gun-teams'_, which - until the test-merge - didn't really exist. I absolutely love the extra roleplay that's spawned from this change, especially when it comes into play during a _'defense'_ mission _(like Whiskey Outpost or Blackstone Bridge)_. Having to actually use short, controlled bursts to conserve your drum's dwindling belt while your buddy's sprinting a hundred meters to fetch a fresh drum like they're an ammo-jockey in the trenches of Verdun? That shit's absolutely **kino**, and I hope it stays.

FslashN:
I've seen this in action a couple of times with the current values. The damage could be bumped up by another 5 to 47. Maybe an increase in fire rate. Maybe both. The SG doesn't feel all that different than shooting with a regular rifle as of this comment. Which isn't all the surprising given the values.

grifp71:
> I've seen this in action a couple of times with the current values. The damage could be bumped up by another 5 to 47. Maybe an increase in fire rate. Maybe both. The SG doesn't feel all that different than shooting with a regular rifle as of this comment. Which isn't all the surprising given the values.


I agree, I'm trying to collect some hard stats but most users seem to think it needs to punch up about five more damage on regular. I think I'd leave AP where it is, at least for now, to not trivialize crushers.

grifp71:
lets try this. increased rate of fire (like 150%), normal vanilla damage values

grifp71:
I am getting a lot of positive feedback with this ROF buff. I'd like to continue testing this but I think this might be a winner.

Ollanius:
> I am getting a lot of positive feedback with this ROF buff. I'd like to continue testing this but I think this might be a winner.

Absolutely.
Played a few more rounds with the tweaked firerate, and it feels perfect. It's powerful enough that a Smartgunner can lock down a hallway _(and hose a couple dozen bugs down in the process)_, but still forces them to be mindful of their ammunition and to cooperate with other drum-carrying squadmates.
I think you've nailed it down, perfectly.

morrowwolf (approved):

PR #214, opened 2024-04-05
Adds the ARC to the epic APC roster

DexterDude:
Adds the ARC to PVE.

Essentially it is just a reskinned APC intended for smaller unit sizes, namely FORECON but I can see it being used with the standard platoon as well. All stats are the same as the APC, with the exception of the frontal cannon which has a higher fire rate in exchange for having no IFF. This ammo can go quick, too.

APC exterior sprites by: Esselnek
APC interior sprites (with the exception of the door) by: Žydras


morrowwolf:
Link the PR you're porting from please

DexterDude:
> Link the PR you're porting from please

This isn't a port. It's essentially just a reskinned APC, nothing is borrowed from the code over on PVP, just the sprites made by esselnek. Unless that still counts as a "port?"

FslashN:
Need someone familiar with vehicles to look over this.

code/modules/vehicles/apc/arc.dm
```diff
@@ -0,0 +1,157 @@
...
+
+/obj/vehicle/multitile/apc/arc/set_muzzle_offsets(obj/item/hardpoint/HP)
+	//sets muzzle flash offsets for APC weapons
+	switch(HP.slot)
+		if(HDPT_SUPPORT) //flare
+			HP.muzzle_flash_pos = list(
+				"1" = list(22, -20),
+				"2" = list(11, 32),
+				"4" = list(-5, 10),
+				"8" = list(37, 24)
```
Doubleumc, line 106:
```suggestion
			HP.muzzle_flash_pos = list(
				"1" = list(-4, -40),
				"2" = list(5, 5),
				"4" = list(-25, 6),
				"8" = list(25, -4)
```
For hardpoints that can be used on multiple vehicles with different mounting positions, you can override their existing muzzle flash position in this proc. You'll need to set the offsets appropriately for the vehicle, however. The existing offsets were correct for the Movie APC, but were wildly out of position of the ARC's flare mounting point. These new offsets are much closer, and can be adjusted from there.

code/modules/vehicles/hardpoints/primary/arc_frontal.dm
```diff
@@ -0,0 +1,39 @@
+/obj/item/hardpoint/primary/arc_frontalcannon
+	name = "Bleihagel RE-RE850 Frontal Rotary Cannon"
+	desc = "A preculiar turret with two rotary cannons strapped to the front of it. Normally equipped with an autonomous mode, this one curiously lacks the feature."
```
Doubleumc, line 3:
```suggestion
	desc = "A peculiar turret with two rotary cannons strapped to the front of it. Normally equipped with an autonomous mode, this one curiously lacks the feature."
```

code/modules/vehicles/hardpoints/primary/arc_frontal.dm
Doubleumc, line 1: This hardpoint currently lacks IFF, is that intentional? If not, you can implement IFF by adding something like this:
https://github.com/PvE-CMSS13/PvE-CMSS13/blob/2b51c95040e12cf3e554fe6f3dab3a3df914d1e2/code/modules/vehicles/hardpoints/secondary/frontal_cannon.dm#L41-L45

Doubleumc, line 1: If the `fire_delay` is adjusted and IFF added, the `arc_frontalcannon` will be identical to the APC `frontalcannon`. In which case, I can't think of a compelling reason not to use the existing `frontalcannon` instead.

Doubleumc (changes requested):

Doubleumc:
Oh and to be clear: once the above are addressed everything LGTM

DexterDude:
Added a weaponless variant for GMs who dont want to deal with guns. Also fixed all the requested changes.

Doubleumc (approved):
LGTM

PR #215, opened 2024-04-05
Movielike Spawning (Satchel on Spawn Removal)

Max-023:
Removes satchels from cryospawn loadouts, as personnel lockers have rendered them unnecessary. Additionally, adds the service cap to the PltCO's loadout to complete the service uniform set and adds the officer cap to the wardrobe.


AmoryBlaine:
The service jackets are a part of the service uniform. The whole ensemble should be in there if you're gonna put the jackets in. You don't see guys walking around with BDUs, combat boots and their service uniform jackets.

Max-023:
Fixed

Max-023:
I'll remove it from the lockers for now then and come up with another way to include them.

Max-023:
The boiler suit palette adjustment is more for the fact that I am trying to make sure this PR remains compatible with PR#173. Adjusting CM_hats.dmi would make it conflict with said PR. Once that PR is merged, the plan was to adjust the suits back and make the caps use the boiler suit palette, rather than the officer service uniform.

morrowwolf:
Conflicts

morrowwolf (changes requested):
If there's nothing in the function any longer you should remove the typepath/proc override.

AndroBetel:
stale for more than a month, author seems to be inactive

PR #219, opened 2024-04-07
armory and door update

AmoryBlaine:
Changes the doors of the Arrow to USCM paletted doors and swaps the armory stock from breakable crates to openable crates.


DexterDude:
Not really a fan of this change. I prefer the white doors, it adds more colors to the ship and makes it more interesting to look at than just dark dark dark and more dark colors. The crate thing is cool - although I really like of the old break open crates too.

bearrrrrrrr:
kind of echoing above post, breaking stuff also allows ppl more stuff to do (making crates/etc)

PR #220, opened 2024-04-07
Empty Rifles in the Gun Racks

WolfOrion:
Adds a subtype for the M41A mk1 and Type 71 and their respective gun racks which has the rifles unloaded and their safeties toggled on. Replaces the gun racks on each shipmap with these subtypes. Does NOT replace the one in the movie APC. The rifles in that rack are still loaded.


code/game/objects/structures/gun_rack.dm
```diff
@@ -48,13 +48,19 @@
...
+	allowed_type = /obj/item/weapon/gun/rifle/m41aMK1/empty
+
 /obj/structure/gun_rack/type71
 	icon_state = "type71"
 	desc = "Some off-branded gun rack. Per SOF and UPPA regulations, weapons should be stored in secure safes and only given out when necessary. Of course, most (but not all!) units overlook this regulation, only storing their firearms in safes when inspection arrives."
 	max_stored = 6
 	initial_stored = 6
 	allowed_type = /obj/item/weapon/gun/rifle/type71
 
+/obj/structure/gun_rack/type71/empty
```
FslashN, line 61: The rest of the code is fine, but `/obj/structure/gun_rack/type71/empty` implies that the rack is **empty**, not that it has unloaded guns. `/obj/structure/gun_rack/type71/unloaded` or something would be best.

FslashN (changes requested):

code/game/objects/structures/gun_rack.dm
```diff
@@ -48,13 +48,19 @@
 /obj/structure/gun_rack/m41
 	allowed_type = /obj/item/weapon/gun/rifle/m41aMK1
 
+/obj/structure/gun_rack/m41/empty
+	allowed_type = /obj/item/weapon/gun/rifle/m41aMK1/empty
```
AndroBetel, line 52: if you somehow get regular, loaded rifles, you won't be able to put them onto the rack. i don't think that's intentional, maybe modify Initialize() proc somehow?

AndroBetel (changes requested):

code/game/objects/structures/gun_rack.dm
FslashN: `if(istype(O, allowed_type) && contents.len < max_stored)` It should store regular rifles fine. The istype() check will check for `/obj/item/weapon/gun/rifle/type71` and `/obj/item/weapon/gun/rifle/m41aMK1`. This returns true for children of those types.

FslashN: You would need a new variable from `allowed_type`, something like `populate_type`. On Initialize(), instead of using `allowed_type` to populate, it would use the new variable. Then you would change things like `allowed_type = /obj/item/weapon/gun/rifle/m41aMK1/empty` to your new variable. So like, `populated_type = /obj/item/weapon/gun/rifle/m41aMK1/empty`.

WolfOrion: Added populate_type as a new variable in my local code and tested it with an unloaded rack - I could put a spawned m41aMK1/ap into it and get it back, so I think it works. Just need to edit my pull request to reflect this.

AndroBetel (approved):

FslashN (approved):

FslashN:
This is good to go. Good changes, easy to implement into existing maps.

PR #221, opened 2024-04-07
Donger SG Sprites

Max-023:
Resprites the SG equipment to actually look good, as well as adds a version of the M70 flak vest that you can attach to your uniform just for the look, now available in the gear section of character creation for smartgunners.

Also adds sprites for a potential smartgunner pistol belt that would go on their back, as well as a new M56 harness that is with a 'movie accurate' palette with plans on implementation.


Ollanius:
d4c882dd2e707783725565e21722591a
c73fcb0c30f964f5755e6dece7dd0c04

code/game/objects/items/storage/large_holster.dm
```diff
@@ -383,3 +383,142 @@
...
+	name = "\improper M277 Pistol Holster Rig"
+	desc = "holster"
+	icon = 'icons/obj/items/storage.dmi'
+	icon_state = "m56b_holster"
+	storage_slots = 5
+	max_storage_space = 11
+	storage_flags = STORAGE_FLAGS_POUCH|STORAGE_ALLOW_QUICKDRAW
+	var/flap = TRUE
+	var/list/obj/item/weapon/gun/holster_slots = list(
+		"1" = list(
```
FslashN, line 397: I wouldn't do this. You can't associate numbers because they serve as the key for the index. But associating a number string with a index is mostly confusing. There is usually a better way to do it.

code/game/objects/items/storage/large_holster.dm
```diff
@@ -383,3 +383,142 @@
...
+			"gun" = null,
+			"underlay_sprite" = null,
+			"underlay_transform" = null,
+			"icon_x" = 0,
+			"icon_y" = 0))
+
+	var/list/holstered_guns = list()
```
FslashN, line 404: Does this need to be a list? Is it possible to fit more than one gun into the holster? If it is, might be fine.

code/game/objects/items/storage/large_holster.dm
```diff
@@ -383,3 +383,142 @@
...
+/obj/item/storage/large_holster/m56/update_icon()
+	overlays.Cut()
+
+	if(content_watchers && flap)
+		return
+	var/magazines = length(contents) - length(holstered_guns)
+	if(!magazines)
+		return
+	if(magazines <= (storage_slots - length(holster_slots)) * 0.5) //Don't count slots reserved for guns, even if they're empty.
+		overlays += "+[base_icon]_half"
```
FslashN, line 430: Can be written as:
`overlays += "+[base_icon]_[magazines <= (storage_slots - length(holster_slots)) * 0.5) ? "half" : "full"]"`

code/game/objects/items/storage/large_holster.dm
```diff
@@ -383,3 +383,142 @@
...
+	if(current_gun)
+		/*
+		Have to use a workaround here, otherwise images won't display properly at all times.
+		Reason being, transform is not displayed when right clicking/alt+clicking an object,
+		so it's necessary to pre-load the potential states so the item actually shows up
+		correctly without having to rotate anything. Preloading weapon icons also makes
+		sure that we don't have to do any extra calculations.
+		*/
+		playsound(src, drawSound, 7, TRUE)
+		var/image/gun_underlay = image(icon, current_gun.base_gun_icon)
```
FslashN, line 502: You can probably use vis_contents here instead of using image. You essentially add the object itself as an underlay for the sprite and manipulate it accordingly.

FslashN (changes requested):
Left a few comments. I didn't look that indepth into it, so think of my comments as more of suggestions than hard requests. Close them as you go through them. I'm sure others may have some as well.

I like that you commented your code, always good to do that.

code/game/objects/items/storage/large_holster.dm
Max-023: I've tried to edit it but things seem to break for some reason.

Max-023:
Most of the code is ripped from the M4A3 pistol belt code, so that's where most of the comments come from. I've just been modifying the code to make it work with what I had in mind. For some reason, though, the onmob doesn't update when you put a pistol in the holster, when it should.

Ollanius:
6a6a5c27999b5a615dec216ab141fa5c
](url)
010ad510e3d69120ca645bac290567e4

Had the opportunity to test out the pull request on Blue.
- The updated Combat Harness is working nicely. If possible, I'd recommend checking the relevant code/icons _(that determine how bloody overlays are applied to it)_, just to make sure that they're accounting for the extra pixels on the sides.

- The Flak Vest accessory looks great. You're currently unable to spawn with it, though: even if you roll as a Smartgunner, the loadout system will think you're in an _'invalid role'_. That should be a simple fix. Alternatively, you could just make it so that the Flak Vest spawns inside of the Smartgunner's lockers, or - if Morrow's fine with it - make it an unrestricted loadout item.

Good work, overall!

code/game/objects/items/storage/large_holster.dm
```diff
@@ -383,3 +383,141 @@
 	if (!istype(FP))
 		return
 	FP.toggle_fuel()
+
+/obj/item/storage/large_holster/m56
+	name = "\improper M277 Pistol Holster Rig"
+	desc = "holster"
+	icon = 'icons/obj/items/storage.dmi'
```
AndroBetel, line 390: needs cool desc

code/modules/clothing/under/ties.dm
```diff
@@ -138,6 +138,11 @@
 				return
 	return ..(M,user)
 
+//Drake Vest
+/obj/item/clothing/accessory/flak
+	name = "M70 flak vest"
+	desc = "Venlar flak jacket worn by combat support personnel such as dropship crew, or occasionally by smartgunners. Despite the name it's actually better at stopping ballistics...."
```
AndroBetel, line 144: misleading desc since it has no armor + too much periods!!!!!

code/game/objects/items/storage/large_holster.dm
```diff
@@ -383,3 +383,141 @@
...
+	var/list/holstered_guns = list()
+
+	var/sheatheSound = 'sound/weapons/gun_pistol_sheathe.ogg'
+	///Used to get flap overlay states as inserting a gun changes icon state.
+	can_hold = list(
+		/obj/item/weapon/gun/pistol,
+		/obj/item/ammo_magazine/pistol,
+	)
+
+/obj/item/storage/large_holster/m56/post_skin_selection()
```
AndroBetel, line 412: does it really need to redefine all these procs..??? am i missing something?

AndroBetel (changes requested):

PR #222, opened 2024-04-07
Medals

morrowwolf:
Adds medals to the overwatch console. SO can give them.

Don't meme with them please.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -9,7 +9,7 @@
 		"**/.pnp.*": true
 	},
 	"editor.codeActionsOnSave": {
-		"source.fixAll.eslint": true
+		"source.fixAll.eslint": "explicit"
 	},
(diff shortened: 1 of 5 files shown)


DexterDude:
Don't know about this one. While cool, I worry constant overuse will devalue the medals. Realistically marines kinda do stuff most ops that WOULD get them medals, but them being in constant circulation by LTs who want to award their men kinda makes them less cool to get.

At best I feel as though they should be given limited medals to give you, like only the bronze heart and MAYBE distinguished conduct. Everything else above should be GM only to retain their value - LTs giving out the equivalent of a medal of honor is kinda ehhh, same with I think the equivalent of the silver star?

FslashN:
I think Dexter brings up some good points. Something to TM, perhaps.

morrowwolf:
Distinguished only now, which is the "basic" medal

PR #225, opened 2024-04-08
Fixes three instances of a negative accuracy bonus

AmoryBlaine:
# About the pull request

Fixes three instances of negative accuracy bonus being applied to the M4A3, 1911 and DEagle generic ammos. Currently the state of the guns affected by this accuracy nerf means you're tossing away a good few rounds per mag for fuckups you have no control over in situations already dire enough to justify the use of a pistol. It makes it unfun to use knowing you're being penalized for using it when it doesn't even really impact the greater aspect of the game. In particular it's bad on the 1911 since you only have 7 rounds anyways


FslashN:
Is this actually a fix? Please test these with numbers, because it may be intentional to the performance of the gun.

bearrrrrrrr:
wtf, this changes all bullets at base
what???

AndroBetel (approved):
vasquez killed a warrior with a pistol

PR #229, opened 2024-04-10
Adds a new shipmap, the USS Heyst

DexterDude:
Adds the USS Heyst to the roster of shipmaps. Instead of having a unique platoon, this shipmap instead offers an alternative for GMs to use should they so desire. The platoon is the exact same as the Golden Arrow and it comes equipped with everything the Golden Arrow has.*

The USS Heyst is a Conestoga class light assault starship used by the Colonial Marines. The most famous ship of this class is the USS Sulaco featured in Aliens for very obvious reasons. The map is based on an amalgamation of the movie, ACM video game, and aliens the blueprints. The core focus of this map is to encourage more RP from the marines pre and post op, removing the squad based armories in favor of a centralized one, as well as placing most of the essential shipside areas in the same general region as one another vs. the Golden Arrow which does have them a bit more spread out.

The USS Heyst is not a ship in proper Aliens lore, however it is in SS13 PVP CM lore as a vague unclassified ship that was involved in Tychon Tackle. Given the lore behind it there, it can safely be considered the same exact ship and thus there aren't any lore conflicts either. There is a lore blurb on the Maximum Effort banner that alludes to this.

*The Heyst does not come with a platoon sergeant's quarters, supply bay, or overwatch. The supply bay is replaced by a cargo lift elevator that can be opened with buttons hidden on the outside of the hull. Platoon sergeant no longer has his own private quarters, he bunks and gears up with his men like he should - though a cramped shared office is accessible to all SGTs for RP purposes. 

Overwatch is nonexistant, instead you are expected to use the Movie APC and deploy the Platoon Commander much like the movies. If you wish for the LT to not technically deploy, you can have them stay in the APC in the dropship, or have them sit around the LZ.

Out of date and mislabeled map attached.

heyst


AndroBetel (changes requested):
add heyst to shipmap list define in code\global.dm

DexterDude:
dnm somethings fucky with the new areas theyre turning into space

DexterDude:
should be good if linters doesnt explode :)

DexterDude:
ok should be good fr

AndroBetel (approved):

FslashN:
I think this should be fine to add into rotations for GMs to select.

PR #231, opened 2024-04-10
M37 is now a proper heirloom

Max-023:
Drops the superbuck M37's internal magazine capacity down to 4 (o bring it in line with the actual movie prop), gives it a stock by default that the player can choose to remove, and changes its name and description to emulate that of a generational heirloom. Drops the case count on the Golden Arrow down to one and puts it in the dormitory, as a first come first serve basis. To compensate for the loss of firepower, an M41E2 has been placed alongside the M4RA in the Sergeant's room that can be given out. Finally, a movielike M37.

For real though, this should hopefully reduce the amount of unintentional battlefield executions by superbuck by half (on paper that is) by making it be more of an irregular weapon a Marine had onboard as a personal carry, like other Marines and their custom sidearms. Dropping the capacity should also help incentivize using a pulse rifle as well, as more frequent reloads would be required now if the shotgun is overutilized. The M41E2, on the other hand, better fits as a limited service weapon that could be issued out to augment a squad's firepower.

It's thematically weird that both squads of the section would be issued heavily modified hunting shotguns with such high capacity despite the weapon supposedly not being in service at all. The Colonial Marines video game is notably conflicting with the movie in the shotgun's representation, as well, so I've tried to bring the M37 on the golden arrow at least closer to how it is in the movie with an optional stock attachment for someone who would want a little more originality.


Ollanius:
I prefer keeping the Shotgun as-is.

It's a squad-level _'power weapon'_, similar to the Incinerator and Sentry: limited ammunition, but extreme damage. Its presence  ensures that everyone in the squad _(barring the Smartgunner)_ has the option to diversify their loadout, if they wish to do so. While there has been a good amount of deaths unintentionally caused by buckshot, I'd hardly cite it as a reason to limit its accessibility. Just like with the Incinerator and Battle Rifle, situational awareness and careful aiming is vital to maximizing the Shotgun's usefulness - and as long as it's accessible by players in _any_ capacity, unintentional deaths will follow.

I wouldn't call the Shotgun's presence thematically weird, either. The Colonial Marines are often deployed to areas where a large-bore weapon would be invaluable: namely, urban environments (rife with close-quarters-combat) and inhabited colonies (teeming with hostile xenofauna). That's not accounting for the general need of door- _(and airlock-)_ breaching tools, as well, which the Shotgun excels at. Even if you discount that, the Ithica _(despite being a 'heavily modified hunting shotgun')_ has already been combat-proven by the Colonial Marines' predecessor in the twentieth century's most brutal conflicts.

It's a limited-issue weapon that - owing to its storied history and utility - has remained in service with the Colonial Marines. It might not be _'movie-accurate'_, but neither is the Battle Rifle or most of the sidearms: and for the sake of variety _(and gameplay)_, I think that's fine.

I could resprite the Shotgun to better fit its depiction in Dark Descent _(with black furniture, an extended magazine tube, and a shell rack)_ if we want to make it look more fitting as a weapon produced by Armat.

Max-023:
I would disagree on the part of it devaluing the Smartguns, since I think you missed the part where I said I put it in the Sergeant's room with the M4RA. That means there is only one and with much more limited belts. And considering the M41E2 cannot go into automatic fire mode without a bipod, and I didn't include a bipod on the map, that means the Smartgun still claims the unique role of being the SAW. And I only added it to compensate for the loss of one of the M37s.

I do agree, the M37 provides an important role roster wise, but I think that is the biggest problem with the weapon. It was a random shotgun Hicks brought with him out of tradition, not because it was a real USCM weapon. Making it so prevalent and pivotal in my eyes detracts from the M41A, the workhorse of the USCM, and the other weapons available that are actually treated as supplied weapons.

grifp71:
I am really not sure how I feel about this because I feel like it incentivizes just bumrushing the dorm rooms to get the shotgun. Right now you're only really in competition against maybe one other person as the squad rifleman if you want the shotgun. It's also been this way for months and honestly the shotgun FF isn't that bad. I feel like it's a part of our identity, but I'd rather see shotgun FF damage get nerfed than the shotgun become rarer, I think, but I really think no change is necessary at all.

I'm 50/50 on the HPR. I like Verbs idea of maybe the MK221 being standard but it's also semi auto and that with superbuck is a nightmare.

AndroBetel:
stale for more than a month, author seems to be inactive

PR #233, opened 2024-04-11
[s] Spacebeast Revived

FslashN:
Builds off the following PR: https://github.com/PvE-CMSS13/PvE-CMSS13/pull/161
Credit to @AndroBetel for the initial sprites and code.

I've added additional logic and behavior, enabled abilities, and mildly refactored some code to make this possible. Details to follow later.


code/modules/admin/holder2.dm
```diff
@@ -105,7 +105,7 @@ you will have to do something like if(client.admin_holder.rights & R_ADMIN) your
 	if(!other)
 		return FALSE
 	if(rights_required && other.admin_holder?.rank)
-		if(check_client_rights(usr.client, rights_required, show_msg))
+		if(check_client_rights(other, rights_required, show_msg))
```
FslashN, line 108: Runtime error fix.

code/modules/mob/living/carbon/human/life/handle_breath.dm
```diff
@@ -8,7 +8,7 @@
 		return
 	if(istype(loc, /obj/structure/machinery/cryo_cell))
 		return
-	if(species && (species.flags & NO_BREATHE || species.flags & IS_SYNTHETIC))
+	if(species && (species.flags & (NO_BREATHE|IS_SYNTHETIC)))
```
FslashN, line 11: No need for two checks. Not really related to the rest of the PR.

code/modules/mob/mob_grab.dm
```diff
@@ -24,8 +24,11 @@
 
 /obj/item/grab/Destroy()
 	grabbed_thing = null
-	if(ismob(loc))
-		var/mob/M = loc
+	var/mob/M = loc
+	if(istype(M))
+		/// If a mob is qdeleted while grabbing, the stack will call for stop_pulling(), which will call for this item to be deleted on drop.
+		/// But it is already being deleted through previous procs in the stack; so we don't want to call the function again when it is dropped.
+		if(M.gc_destroyed == GC_CURRENTLY_BEING_QDELETED) flags_item &= ~DELONDROP
```
FslashN, line 31: Runtime fix.

Ollanius:
https://www.youtube.com/watch?v=8je4bO3MVc8

I've recorded a round that predominately features the Spacebeasts.

The increased lethality is nice, but it feels like they're a bit too resilient - at least in normal circumstances. Their abilities _(such as tailsweeping)_ aren't very pleasant to fight against: I feel like stun-based abilities should be restricted to special Xenomorphs _(like the Lurker or Runner)_, or else you'll have situations where a lone marine will get mercilessly stun-juggled into oblivion. I've had a couple instances where Spacebeasts were also able to pass straight through me _(despite having 'harm intent' selected)_, which was a bit jarring.

Fairly enjoyable to fight, either way. If they're going to be more prominent, however, I'd recommend removing the 'tailsweeping' ability.

FslashN:
> https://www.youtube.com/watch?v=8je4bO3MVc8
> 
> I've recorded a round that predominately features the Spacebeasts.
> 
> The increased lethality is nice, but it feels like they're a bit too resilient - at least in normal circumstances. Their abilities _(such as tailsweeping)_ aren't very pleasant to fight against: I feel like stun-based abilities should be restricted to special Xenomorphs _(like the Lurker or Runner)_, or else you'll have situations where a lone marine will get mercilessly stun-juggled into oblivion. I've had a couple instances where Spacebeasts were also able to pass straight through me _(despite having 'harm intent' selected)_, which was a bit jarring.
> 
> Fairly enjoyable to fight, either way. If they're going to be more prominent, however, I'd recommend removing the 'tailsweeping' ability.

Thanks for the feedback. Do you happen to have a timestamp in the video for when the alien was able to pass through you? That shouldn't be happening.

messagecypher:
don't like how they're always faster than you at around lurker-level of speed while being tougher
combined with the fact that they can lunge you, smack you, and absolutely ruin your aim

soldiers seem to be relatively overpowered - if you don't see them coming and you aren't shooting at them the moment they enter your screen, it's going to reach you and absolutely debilitate you which is a death sentence on fast-paced situations in PVE
not to mention the fact that their warrior punch and tackle-spam means they're capable of easily taking out a marine, RNG willing

tl;dr soldier speed and grab make fighting back too hard and really easy for everything else to kill you (including the soldier)

code/modules/mob/living/carbon/human/human_helpers.dm
```diff
@@ -198,26 +198,26 @@
 	if(istype(brain))
 		return TRUE
 	return FALSE
-
+f
```
xDanilcusx, line 201: funny letter

code/modules/mob/living/carbon/xenomorph/ai/xeno_ai.dm
```diff
@@ -34,6 +34,9 @@
 	/// Cooldown declaration for delaying finding a new path if no path was found
 	COOLDOWN_DECLARE(no_path_found_cooldown)
 
+	var/ai_active_intent = INTENT_HARM
+	var/target_unconscious = FALSE
```
xDanilcusx, line 38: I don't know if we really need something like this var, because currently we have ``ai_check_stat()`` to clarify which stat we need to continue targeting something. It's simple and 
For reference, check [ `/mob/living/carbon/human/ai_check_stat(mob/living/carbon/xenomorph/X)`](https://github.com/PvE-CMSS13/PvE-CMSS13/blob/1d6758c237cf302e32fc0aaa7ece76e8f60fa716/code/modules/mob/living/carbon/xenomorph/xeno_ai_interaction.dm#L181)

Here i made a check to allow facehuggers to target humans even when they are uncon

FslashN: I would really like to avoid typechecking as much as possible. The purpose of `target_unconscious` is to provide a boolean behavior gate that can be repeated for multiple types of aliens. I'm fine with making a specific case for `human/ai_check_stat()`, but I'm not sure typechecking is the way to go. Making it into a variable, on the other hand, seems pretty elegant. Though I would of preferred it as a flag instead of a boolean.

Ollanius:
> Thanks for the feedback. Do you happen to have a timestamp in the video for when the alien was able to pass through you? That shouldn't be happening.

No problem - I really like this 'Soldier', by the way. Just needs a bit more tweaking, in my opinion.
Here's a timestamped link that shows it off.

https://youtu.be/8je4bO3MVc8?si=EsjjzRbXyG882rLv&t=169

code/modules/mob/living/carbon/human/human_helpers.dm
FslashN: Yuuup. It got lost.

Merrgear:
The soldier right now feels like its trying to be both a harasser and a mainline xenomorph, both of these combined isnt very fun to fight. IMO its too fast for the amount of tank it can take and the disarm spam and grab spam isnt very fun to fight against. This is part of why i fundamentally disagree with Amory's movie accurate drone idea, but atleast he thinks of movie accurate as not tanking anything. Part of why i play PVP is to be free from the annoying to fight tactics like said disarm spam. I could get down with it MAYBE if it had a much lower chance to knock down and couldnt eat 12 bajillion rounds to the teeth if you are lucky enough to get up.

But all of that in mind, it still feels like treaded ground, the stun warfare and speed of a cloaked lurker is already done BY the runner and lurker, maybe if it had more abilities centralized around being more like drones and less like a harasser that somehow can eat 10 rounds.

tldr: Too fast, too tanky, stun warfare is unfun to fight against

code/modules/admin/game_master/game_master.dm
```diff
@@ -29,7 +29,7 @@ GLOBAL_VAR_INIT(radio_communication_clarity, 100)
 
 // Spawn stuff
 #define DEFAULT_SPAWN_XENO_STRING XENO_CASTE_DRONE
-#define GAME_MASTER_AI_XENOS list(XENO_CASTE_DRONE, XENO_CASTE_RUNNER, XENO_CASTE_LURKER, XENO_CASTE_CRUSHER, XENO_CASTE_FACEHUGGER)
+#define GAME_MASTER_AI_XENOS list(XENO_CASTE_DRONE, XENO_CASTE_SOLDIER, XENO_CASTE_RUNNER, XENO_CASTE_LURKER, XENO_CASTE_CRUSHER, XENO_CASTE_FACEHUGGER)
```
Doubleumc, line 32: You likely want to add it to the Ambush menu as well:
https://github.com/PvE-CMSS13/PvE-CMSS13/blob/bf5640215da02fdd6f420b9d2f560b70bebb2d34/code/modules/admin/game_master/game_master_submenu/ambush.dm#L3

private-tristan:
Soldiers don't stand still during headbite so most of the time they end up dancing around their target during the headbite windup and failing it (outside of times where they walk back into the tile where they started the headbite).

maybe make headbite root the soldier and give it a longer windup?

private-tristan:
from what i've seen, soldiers move too quickly and have too much HP, to the point where there's no way for marines to feasibly kill a soldier (outside cades) before it wombo combos and does an absurd amount of damage to a single marine. They either need to be slower, have less hp, or (as amory said) have footsteps sound so that marines can deal with them.

WolfOrion:
Would it be possible to make their capture behavior dependent on capture markers placed on individual marines rather than on hive markers? They often interfere with other xenomorphs that capture marines using markers since soldiers follow their own logic separately from capture markers.

PR #235, opened 2024-04-12
Adds MOPP M2 for ROLEPLAY

DexterDude:
MOPP M2. Visually it is the same as MOPP M3 but it's able to be worn by marines under their armor and with a helmet. Has no mechanical differences from standard marine uniforms. Instead of being a hat the M2 mask goes in the mask slot.

mopprine attached

mopprine

### DIFF
--- a/code/modules/clothing/masks/gasmask.dm
+++ b/code/modules/clothing/masks/gasmask.dm
@@ -31,6 +31,17 @@
 	icon_state = "kutjevo_respirator"
 	item_state = "kutjevo_respirator"
 
+/obj/item/clothing/mask/gas/pve_mopp
+	name = "M2 MOPP mask"
+	desc = "The M2 MOPP mask includes a full covering cowl that securely attaches to the MOPP suit. It is capable of protecting of a variety of radiological and biological threats."
+	icon = 'icons/obj/items/clothing/cm_hats.dmi'
+	icon_state = "cbrn_hood"
+	item_state = "cbrn_hood"
+	flags_inv_hide = HIDEEARS|HIDEFACE|HIDEALLHAIR
+	item_icons = list(
+		WEAR_FACE = 'icons/mob/humans/onmob/head_1.dmi'
+	)
+
 /obj/item/clothing/mask/gas/pmc
 	name = "\improper M8 pattern armored balaclava"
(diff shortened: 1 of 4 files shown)


AmoryBlaine:
Can you make it so it attaches to the uniform so we can take it off mid-mission rather than wear it as its own uniform

DexterDude:
> Can you make it so it attaches to the uniform so we can take it off mid-mission rather than wear it as its own uniform

i dont want it getting fucky with layers like webbing and such tbh.

code/modules/clothing/masks/gasmask.dm
```diff
@@ -31,6 +31,17 @@
 	icon_state = "kutjevo_respirator"
 	item_state = "kutjevo_respirator"
 
+/obj/item/clothing/mask/gas/pve_mopp
+	name = "M2 MOPP mask"
```
Zonespace27, line 35: `\improper`

Zonespace27 (changes requested):

code/modules/clothing/masks/gasmask.dm
```diff
@@ -70,6 +81,11 @@
 	icon_state = "bear_mask"
 	anti_hug = 2
 
+/obj/item/clothing/mask/gas/pmc/marsoc
```
AndroBetel, line 84: thats not mentioned in PR body, what's this for again?

code/modules/clothing/under/marine_uniform.dm
```diff
@@ -1042,6 +1042,20 @@
 	icon_state = "rmc_uniform_lt"
 	worn_state = "rmc_uniform_lt"
 
+/obj/item/clothing/under/marine/pve_mopp
+	name = "\improper M2 MOPP suit"
+	desc = "M2 MOPP suits are purpose built to defend the wearer against biological and radioactive contaminants, from nerve gas to nuclear fallout."
+	desc_lore = "The several-paragraph long expository pamphlet that usually comes with these is missing."
```
AndroBetel, line 1048: funny...

DexterDude: it is

AndroBetel:
awesome

PR #237, opened 2024-04-13
CORSAT palette swap and classic replaced with jungle

AmoryBlaine:
Classic palette exists largely because people were very averse to losing the look of the marines prior to the introduction of jungle. It doesn't actually reflect anything other than this. It's simply a holdover from an era that was not willing to accept a more accurate representation of the USCM as per the movie. I don't see much of a reason not to replace it with jungle given this is more fitting for the marines in those environments given what we see in the movie.


Ollanius:
I disagree with removing the classic palette. It's not only stylish, but also an iconic part of the server's aesthetics.
It's also a pretty sensible choice for when the Colonial Marines are going to an orbital station, such as Fiorina or CORSAT. It wouldn't make much sense to wear camouflage intended for a jungle up there.

With that being said, however, I absolutely agree that the classic palette shouldn't be used on jungle-themed maps _(like LV-624)_, and should be replaced with the appropriate camouflage.

AmoryBlaine:
> It's also a pretty sensible choice for when the Colonial Marines are going to an orbital station, such as Fiorina or CORSAT. It wouldn't make much sense to wear camouflage intended for a jungle up there.
It's a bright teal armor colour with khaki uniform worn aboard dark grey or off-white stations. It doesn't work any better than green. The only difference is that green is what we know they wear. Just because when the OG sprites were made they were made with teal and tan does not mean we should still be using it today. Especially this far detached from it.

Ollanius:
> It's a bright teal armor colour with khaki uniform worn aboard dark grey or off-white stations. It doesn't work any better than green. The only difference is that green is what we know they wear. Just because when the OG sprites were made they were made with teal and tan does not mean we should still be using it today. Especially this far detached from it.

Completely understandable. I simply believe it looks better in those environments - at least, compared to the jungle-camouflaged versions.

FslashN:
What maps still use the classic camo with these changes?

AmoryBlaine:
None, it removes them from rotation entirely. CORSAT is now jungle, Chance,
the two prison maps, and the Almayer are all jungle. The only thing that
isn't jungle is any default overrid maps without the camo input. I might
see how to change that tomorrow. Ideally the only place classic shows up is
if it is expressly desired. Otherwise it's jungle, desert or cold weather.

Doubleumc (approved):
Looks good to me.

PR #240, opened 2024-04-17
Conestogafication of the Golden Arrow

DexterDude:
Makes my conestoga shipmap the default shipmap as per the request of Morrow and renames it to the Golden Arrow. The original USS Golden Arrow is still available to GMs as a "classic" version and it's marked as such.

Also adds some changes to the conestoga as well, namely:
- Removes the Movie APC from the vehicle bay and associated modules (Requested by morrow)
- Adds an overwatch area for the PlatCo, now where the shared office once was (Requested by morrow)
- Moves the shared office to the synthetic closet
- Moves the synthetic closet to the maintenance deck
- Adds a supply bay to the maintenance deck so SOs can send down crates (this was a feature on the OG Arrow too that nobody used! one of the consoles next to the pad is secretly a supply pad console)
- Changes back from Corsat plating to normal plating since it looked odd when the DS was landed.
- Some misc. changes that aren't worthy of noting really


DexterDude:
yea idk i fucked this up somehow and im just gonna re-pr it

PR #242, opened 2024-04-19
Some Golden Arrow changes

DexterDude:
replaces the fuel tank with a static one in the maint deck
removes random sleeper on the medbay table
adds numbered signs to signify what closets/equipment belongs to what squad
added ARES consoles to the marine common room and platoon commander's quarters for mostly RP purposes, but would be an interesting way to communicate with the GM beyond calling.
 
Related, also made it so that O-1 to O-3 cannot call self destruct through the ARES interface.


AndroBetel (changes requested):
while youre at it please fix landing lights k thx

DexterDude:
> while youre at it please fix landing lights k thx

true

AndroBetel (approved):
awesome

PR #248, opened 2024-04-26
Pizza Galaxy PR

DexterDude:
doors are buggy for some reason, they will disappear for a split second when they finish animating. for now they are replaced with generic white doors.

the half walls for this are unused as the layering with them is super fucky, replaced with different hand rails

PR with the sprites: https://github.com/cmss13-devs/cmss13/pull/5963
PR for the station: https://github.com/cmss13-devs/cmss13/pull/5887

Map by Nanu308, modified slightly by me
Sprites by wei/esselnek, given approval to PR them despite not being on pvp


DexterDude:
TMed and worked fine

FslashN:
If the doors are behaving strangely, look into `openspeed`. It may not be set correctly; it should coincide for how long their animation actually takes place.

Drulikar:
> If the doors are behaving strangely, look into `openspeed`. It may not be set correctly; it should coincide for how long their animation actually takes.

Its literally one of the reasons why the PR being ported here isn't merged. I'd recommend just waiting for upstream to have the station complete first before porting it.

PR #257, opened 2024-05-07
Tank weapon IFF

Doubleumc:
# About the pull request

Adds IFF to every tank hardpoint that can take it. The only exception is the secondary flamer, which doesn't work in a way that would allow IFF (not that it would make sense regardless).

Note that this only applies to the *projectiles*, the explosions/flames they cause are indiscriminate.

If a GM wants to use a tank, it should be able to operate in close coordination with the marines without the risk of major friendly fire. The complexities of a large, clumsy box needing to deal with clear firing lanes is more a PvP balance thing, rather than something we'd want to keep for PvE.

<details>

</details>

# Changelog

:cl:
qol: all tank hardpoints have IFF (besides secondary flamer)
/:cl:


DexterDude:
ily

PR #261, opened 2024-05-11
(mostly)Ignorable weeds

Doubleumc:
# About the pull request

Removes base xeno speedup from weeds. Fast resin and abilities like resin-walker still work.
Remove AI xeno healing from weeds. AI xenos can still gain plasma and player xenos can still heal.
Melee weapons now do a lengthy channeled action to clear weeds and nodes. Other removal methods (fire, explosions) still work.

Marines clearing weeds is objectively the right choice, because of the drawbacks they face and the benefits the xenos get from them. While it makes sense from a game mechanics perspective, those mechanics aren't really important to PvE and its such a weird thing to do in-character.

This PR reduces the need to clear or spread weeds and makes them more of a pain to remove, hopefully causing the marines to largely ignore it and get on with doing marine shit.

AI xenos still gain plasma so that drones can still build when ordered.

Player xenos still heal on weeds as befits their special/boss role. This exception I'm uncertain about. On the one hand it allows a xeno player to use their experience unchanged, on the other hand its one more special exception to remember.

Melee weapons no longer apply damage directly to weeds, instead they start a channeled action with a duration based on the weed health versus the weapon damage. As an example, a marine clearing a normal resin node with a boot knife takes 2 seconds. This is hopefully just long enough to discourage idly clearing weeds without being obnoxious.

<details>

</details>

# Changelog

:cl:
balance: weeds no longer heal AI xenos, need a channeled action to remove
/:cl:


DexterDude:
I feel like we would still have to contend with what is basically gut instinct on the Marine's end. I figure they will still keep clearing weeds if theyre able to regardless of whether or not they provide actual benefits to xenos.

private-tristan:
note that this means that marines have no reason to interact with critted xenos (other than when going for most xeno kills), with no risk of them ever healing out of crit on weeds.

Doubleumc:
> I feel like we would still have to contend with what is basically gut instinct on the Marine's end. I figure they will still keep clearing weeds if theyre able to regardless of whether or not they provide actual benefits to xenos.

I agree that clearing weeds is so ingrained they'd probably still do it reflexively. I'm not sure how to tackle that beyond just giving it time to become common knowledge.

I thought about making the weeds non-interactable but there are just so, so, SO many other things tied to weeds that I'm sure there will eventually be a corner case where they'd need to be able to clear weeds. And if they wanted to clear it for cosmetic reasons all power to them.

Doubleumc:
> note that this means that marines have no reason to interact with critted xenos (other than when going for most xeno kills), with no risk of them ever healing out of crit on weeds.

That's correct. While that wasn't an intent of this PR I believe that is still a beneficial change. The downed xeno is already done, beaten, to be dealt with or ignored as the marine desires. I don't think the possibility of a critted xeno reviving adds anything to the game, its just busywork or an interruption of post-combat RP.

DexterDude:
> > I feel like we would still have to contend with what is basically gut instinct on the Marine's end. I figure they will still keep clearing weeds if theyre able to regardless of whether or not they provide actual benefits to xenos.
> 
> I agree that clearing weeds is so ingrained they'd probably still do it reflexively. I'm not sure how to tackle that beyond just giving it time to become common knowledge.
> 
> I thought about making the weeds non-interactable but there are just so, so, SO many other things tied to weeds that I'm sure there will eventually be a corner case where they'd need to be able to clear weeds. And if they wanted to clear it for cosmetic reasons all power to them.

Yeah, fair enough. Slightly unrelated and it would need spritework but I'd love to see skeletal weeds that aren't able to be interacted with.

bearrrrrrrr:
Not a fan of this. Marines already go incredibly fast. Making it even easier for them to speedrun isn't good, as a GM.

bearrrrrrrr:
> I feel like we would still have to contend with what is basically gut instinct on the Marine's end. I figure they will still keep clearing weeds if theyre able to regardless of whether or not they provide actual benefits to xenos.

Just had a round. Marines went way too fast and stopped clearing weeds entirely. Really hugely not a fan of this. It makes them even stronger and makes the world feel way less responsive. Marines can also easily outrun xenos at any point now, which also feels terrible as GM and makes me have to micro it. Ditto w/ speedups and sticky resin.

Honestly, I think I'd probably like it a lot more if drones could still catch marines, which they can't atm.

DexterDude:
> In general for making removing weeds tougher/invulnerable another concern is that drones can spread weeds SO easily and do so whenever idle. Being crowded out by ever-growing weeds doesn't sound fun.

Yeah. Drones having a toggle for weeds would be nice.

bearrrrrrrr:
Yeah, I dunno. It's a difficult design issue.

I don't rightly know how to respond, honestly. The points you bring up are all very good and important lol.

AmoryBlaine:
I agree that weeds shouldn't be destroyable but they absolutely should apply a slowdown still. It's an environmental hazard presently and that should persist.

private-tristan:
For a bandaid solution can't you just make it so that GMs can toggle whether weeds do slowdown/heal drones or are breakable?

Doubleumc:
It would be possible to make a toggle but since we both acknowledge its not an actual solution, as a temporary band-aid I feel it would be confusing rather than helpful. Whether its testmerged or not, whether its toggled or not, whether the player realizes this or not, all leads to a wildly different experience.

Ultimately I'd still like to find a solution that fulfills all the conditions (don't bother to break weeds, weed slowdown, can still handle edge-cases of weed removal) but haven't thought of one yet.

Doubleumc:
Trying something different. Marine slowdown is back in, but now melee weapons take more time to clear weeds.

Staykeu:
> Trying something different. Marine slowdown is back in, but now melee weapons take more time to clear weeds.

marine slowdown shouldn't be paired with the slower time to root weeds, this is just a net disadvantage

i disagree that removing weed slowdown makes it so marines can speedrun ops, marines are going to remove weeds either way

plus, GMs can always make up ways to slow the marines down when they're going too fast, making it so it slows marines down even more by them having to take time to remove it seems like refuge for the odd unimaginative GM 

in my perfect world, i'd have it so the weeds were exactly the same as they are in base, minus the slowdown, that way marines can traverse through weeded areas like the movie without rooting them

i get that marines will have the want to remove weeds ingrained in their brain, but i feel like they'll lose it over time once it's well known that weeds don't slow you down

that's my two-cents as a regular player, take it or leave it hoss, up to you

code/modules/cm_aliens/weeds.dm
```diff
@@ -372,26 +372,29 @@
...
 			playsound(loc, 'sound/items/Welder.ogg', 25, 1)
+			user.animation_attack_on(src)
+			take_damage(WEED_HEALTH_STANDARD)
 	else
+		if(!attacking_item.force)
+			to_chat(user, SPAN_WARNING("You scrape ineffectively at \the [src] with \the [attacking_item]."))
+			return
+		to_chat(user, SPAN_NOTICE("You start clearing \the [src] away with \the [attacking_item]..."))
+		var/duration = (20 SECONDS) * (health / attacking_item.force) //effectively applying attack damage to weed's health spread over 20 seconds
+		if(!do_after(user, duration, INTERRUPT_ALL|BEHAVIOR_IMMOBILE, BUSY_ICON_BUILD))
```
Zonespace27, line 392: this has the do_after wheel keep going even if the user moves, from a round i played in

code/modules/cm_aliens/weeds.dm
```diff
@@ -372,26 +372,29 @@
...
-		to_chat(user, SPAN_WARNING("You cut \the [src] away with \the [attacking_item]."))
-
-	var/damage = attacking_item.force / 3
-	playsound(loc, "alien_resin_break", 25)
-
 	if(iswelder(attacking_item))
 		var/obj/item/tool/weldingtool/WT = attacking_item
 		if(WT.remove_fuel(2))
-			damage = WEED_HEALTH_STANDARD
+			if(istype(src, /obj/effect/alien/weeds/node))
```
Zonespace27, line 378: you don't need \the in any of this (ik it's not your code but still)

Zonespace27 (dismissed):

cm13-github:

cm13-github:

code/modules/cm_aliens/weeds.dm
Doubleumc: Fixed.

code/modules/cm_aliens/weeds.dm
Doubleumc: Unable to replicate. In testing the wheel vanishes after moving.

PR #266, opened 2024-05-20
Garrow Modifications 2

DexterDude:
Fixes some issues, namely some layering stuff and some missing pipes, as well as the platco phone category (thanks andro for reminding me)

Also adds some QOL for GMs by giving them some mostly empty cargo bays and more accessible and clear buttons.

Also adds support:tm: for Andro to add elevators. I looked at it and it seemed a little more complicated than normal shuttles, so I figured I'd leave it to the experts. Don't merge before then, please!


DexterDude:
fail...oh my science...

AmoryBlaine:
image
Gotta fix the layering here

Steelpoint:
Rogue blast door tile at south-west end of the main hanger deck. 

Might be a missile pipe at the far, far, north end of hanger.

code/datums/shuttles.dm
```diff
@@ -125,3 +125,16 @@
 
 /datum/map_template/shuttle/trijent_elevator/ice_elevator/requisitions
 	elevator_network = "Requisitions"
+
```
Doubleumc, line 128:
```suggestion
/datum/map_template/shuttle/golden_arrow
	shuttle_id = "unused"

```
Need a dummy parent type here to avoid annoying the unit tests. `shuttle_id` doesn't matter besides it not being `"SHOULD NEVER EXIST"`.

A subtype implies the existence of its parent type. So a `/datum/map_template/shuttle/golden_arrow/vehicleone` implies that a `/datum/map_template/shuttle/golden_arrow` exists. Since its not explicitly defined, it inherits all its values directly from `/datum/map_template/shuttle`, including `shuttle_id = "SHOULD NEVER EXIST"`.

One of the unit tests checkss all the subtypes of `/datum/map_template/shuttle` to ensure that `shuttle_id` is set to something other than `"SHOULD NEVER EXIST"`. It was triggering on this implied type.

Doubleumc (changes requested):

cm13-github:

cm13-github:

cm13-github:

cm13-github:

code/game/machinery/doors/poddoor.dm
```diff
@@ -298,6 +298,10 @@
 		/obj/structure/machinery/door/airlock,
 	)
 
+/obj/structure/machinery/door/poddoor/almayer/LateInitialize()
+	. = ..()
+	relativewall_neighbours()
```
Doubleumc, line 303: What is the purpose of this change? I'm unable to see the difference it makes.

code/modules/shuttle/shuttle.dm
```diff
@@ -637,6 +640,8 @@
 /obj/docking_port/mobile/proc/on_prearrival()
 	if(destination)
 		destination.on_prearrival(src)
+		var/obj/docking_port/stationary/dropzone = destination
+		playsound(dropzone.return_center_turf(), landing_sound, 60, 0)
```
Doubleumc, line 644: Sounds start playing immediately on landing instead of during the windup to landing. Perhaps `on_ignition()` would be better, or check how the dropships handle it:
https://github.com/cmss13-devs/cmss13-pve/blob/6c5ba9ac4b19f8597978ad5b4399213a3f07ac94/code/modules/shuttle/shuttles/dropship.dm#L212-L227

code/modules/shuttle/shuttles/shipmap_elevator.dm
```diff
@@ -0,0 +1,73 @@
...
+	for(var/i in gears)
+		var/obj/structure/machinery/gear/G = i
+		G.stop_moving()
+	for(var/i in railings)
+		var/obj/structure/machinery/door/poddoor/railing/R = i
+		if(R.density)
+			INVOKE_ASYNC(R, TYPE_PROC_REF(/obj/structure/machinery/door, open))
+		else
+			INVOKE_ASYNC(R, TYPE_PROC_REF(/obj/structure/machinery/door, close))
+
```
Doubleumc, line 48: A leaving elevator's docking port creates a message of it falling into the shaft:
image
image

Suggest adding something like this, which is how the ASRS elevator prevents that issue:
```suggestion
/obj/docking_port/mobile/shipmap_elevator/visible_message() //Prevents message spam with empty elevator shaft - "The empty space falls into the depths!"
	return

```

maps/shuttles/garrow_vehicle_elevator_two.dmm
```diff
@@ -0,0 +1,88 @@
+//MAP CONVERTED BY dmm2tgm.py THIS HEADER COMMENT PREVENTS RECONVERSION, DO NOT REMOVE
+"a" = (
+/turf/open/floor/plating/plating_catwalk,
+/area/shuttle/vehicle_elevator)
+"f" = (
+/obj/structure/machinery/computer/shuttle/elevator_controller{
+	shuttleId = "garrow_vehicle_elevator_two"
+	},
```
Doubleumc, line 8: There's a controller in the middle of elevator two, but not elevator one.

maps/map_files/golden_arrow/golden_arrow.dmm
Doubleumc, line 1: Lower level of the elevators shouldn't have the open/empty turf you can fall in to, since they are the bottom.

Upper open/empty turf should fall into its lower counterpart, instead of into the ASRS pit.

code/modules/shuttle/shuttles/shipmap_elevator.dm
```diff
@@ -0,0 +1,73 @@
...
+	..()
+	for(var/i in gears)
+		var/obj/structure/machinery/gear/G = i
+		G.stop_moving()
+	for(var/i in railings)
+		var/obj/structure/machinery/door/poddoor/railing/R = i
+		if(R.density)
+			INVOKE_ASYNC(R, TYPE_PROC_REF(/obj/structure/machinery/door, open))
+		else
+			INVOKE_ASYNC(R, TYPE_PROC_REF(/obj/structure/machinery/door, close))
```
Doubleumc, line 47: Currently railings don't raise until several seconds after the elevator has left. Perhaps in `on_ignition()` instead?

Doubleumc (changes requested):
My main hangup with this PR is that its modifying internals in how docking ports and shuttles work. I'd like to find a way to avoid that if possible.

First thought was to repath the new elevators to be subtypes of `trijent_elevator`. That would make all the internal changes unnecessary. Issue there is trijent elevators expect doors to control, and the vehicle elevators have none. Could change *that* bit of code instead, which is a lesser change but still a change.

Second thought is that the vehicle elevators in particular are trying to combine some function of trijent-style elevators and ASRS-style elevators, which work differently. That's where the "falls into the depths" message comes from, trijent doesn't use those fall-through turfs like the ASRS does, and the open/empty space is still set to throw you into the ASRS pit. Perhaps the vehicle elevators would work better as ASRS-like effects than full shuttles.

Once conflicts and the other issues below are resolved, this is fine for TMing. Still thinking on how to handle the above before merging -- just not familiar enough with shuttles or the ASRS.

DexterDude:
> Second thought is that the vehicle elevators in particular are trying to combine some function of trijent-style elevators and ASRS-style elevators, which work differently. That's where the "falls into the depths" message comes from, trijent doesn't use those fall-through turfs like the ASRS does, and the open/empty space is still set to throw you into the ASRS pit. Perhaps the vehicle elevators would work better as ASRS-like effects than full shuttles.

My reasoning for having them be elevators is because its much more interesting for the players to bring them onto the elevators themselves and lift them up on their own versus it being a magic vehicle lifting system. That, and it means we dont have to code in extra things like GMs having to enable or disable certain vehicles being able to come up or down. It adds more depth to the map itself without requiring additional code. I agree with mostly everything else however. Eventually when Andro gets around to it the elevators will have little remote controls people can pick up, but for now they can use the call buttons.

maps/map_files/golden_arrow/golden_arrow.dmm
DexterDude: True.

maps/shuttles/garrow_vehicle_elevator_two.dmm
DexterDude: Woops! Forgot to delete that after testing.

cm13-github:

code/modules/shuttle/shuttle.dm
AndroBetel: changed it to play on ignition, now sounds as intended

code/modules/shuttle/shuttles/shipmap_elevator.dm
AndroBetel: fixed

code/game/machinery/doors/poddoor.dm
AndroBetel: looks like it just removes weird timered proc with nicer LateInitialize

Doubleumc: Cleaner looking code to be sure, but would prefer this be done on PvP then pulled here to keep compatibility.

cm13-github:

cm13-github:

maps/map_files/golden_arrow/golden_arrow.dmm
Doubleumc, line 1: What's the purpose of the railing control button for the cargo elevator? The railings are working correctly so they're raised when its unsafe and lowered when its safe.

Doubleumc (changes requested):
Just a few things to tidy up. Besides that, code good.

maps/map_files/golden_arrow/golden_arrow.dmm
DexterDude: Didnt know they automatically changed.

Doubleumc (approved):

Doubleumc (dismissed):

PR #267, opened 2024-05-21
Buildmode & GM panel compatability

Doubleumc:
# About the pull request

Allows multiple click intercept sources (ex/buildmode, GM panel, rappel panel) to coexist peacefully. It remembers all sources and the most recent source takes priority.

Instead of a single `click_intercept`, there's now a list of `click_intercepts`. Any activated intercept source will add itself to the list, and will remove itself when deactivated. The last source in the list (the most recent) gets the clicks.

Significantly adjusted how the GM panel handled its multiple intercepts to work with the new system. Should be the exact same for the user.

Buildmode and the GM panel are used every round, frequently at the same time. This PR allows them to be used side-by-side in a much smoother way than one nuking the other's click intercept if that control is closed.

Boots without issue. Tested having buildmode, GM panel, rappel panel all open at the same time. Closing one allows the others to still function.

# Changelog

:cl:
qol: buildmode & GM panel don't stomp on each other's clicks
/:cl:


cm13-github:

Doubleumc:
Not remapped, I forgot about them while implementing this. The GM menu needs to "own" the click while middle-clicking for those to work, but this changes it to only "owning" the click when one of the spawn/objective/etc on-click toggles is enabled.

No current plans for this. It never became a big issue that needed solving, and I'm not sure this was the best way to solve it regardless (assuming I fixed the middle-click).

PR #272, opened 2024-05-28
rare cigs

Meatstuff882:
# About the pull request

Removing Calcutta Emerald Greens from the loadout menu. Felt inconsistent that those were available but WY Golds, Arcturan Aces, and Lady fingers were the same price or lower but not available. Asked Morrow about it and he said "rare things should be rare". I'm taking this to mean that Calcutta's being available is against that.

I tested it by filling a loadout menu with a bunch of things and calcuttas also, and removing the calcuttas from the loadout menu just erases it from existing loadouts. Didn't seem to cause any weirdness or erasure of other objects.

If other $30 and $35 packs that are available in the nice vendor but not the ship vendor are "rare", then ALL $35 and $30 packs that are in the nice vendor but not the ship vendor should be rare.

Consistency, stimulating the nerves in my brain, encouraging me to make more PRs in the future.

# Changelog

:cl:
del: removed calcuttas from the loadout menu
\:cl:

### DIFF
--- a/code/modules/client/preferences_gear.dm
+++ b/code/modules/client/preferences_gear.dm
@@ -905,10 +905,6 @@ var/global/list/gear_datums_by_name = list()
 	path = /obj/item/clothing/mask/cigarette/cigar
 	cost = 2
 
-/datum/gear/smoking/pack_emerald_green
-	display_name = "Pack Of Emerald Greens"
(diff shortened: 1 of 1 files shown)


DexterDude:
cig equality.

PR #287, opened 2024-06-08
glasses slot doesn't suck anymore

Ediblebomb:
# About the pull request
Moves layers around to make glasses actually look good in their dedicated slot.
Glasses won't look like ass in their own slot anymore.
<details>

image
image

</details>

# Changelog
:cl:
fix: Makes glasses not look like shit in the glasses slot. Is this a fix? I dunno but it's not like these show up in game anyway.
/:cl:


AndroBetel:
yeah no, i think it actually removes variability, since you can put most of glasses into mask slot with same results
it also looks kind of goofy with some hair styles and glasses
image

PR #288, opened 2024-06-08
Vehicle crashing tweaks

Doubleumc:
# About the pull request

Corrects a few math issues and loosens tolerances so vehicle-crash-interior-flinging can actually happen now.

`move_momentum` can be positive for forward or negative for backward, so `Floor(move_momentum/2)` made forward crashes less intense and backward crashes more intense. Corrected to `trunc(move_momentum/2)`, which rounds towards zero consistently.

The maximum value for `move_momentum`, fittingly named `move_max_momentum`, is 2 for the APC and 3 for the tank. Because its divided by 2 and then floored, the max momentum actually fed to the `interior_crash_effect()` proc was 1* (3/2 = 1.5, floor(1.5) = 1). `interior_crash_effect()` has an early return for `if(abs(move_momentum) <= 1)`. Corrected to `if(abs(move_momentum) < 1)` so values of 1 still trigger the crash.

*except the tank moving backwards, (-3/2 = -1.5, floor(-1.5) = -2)

Correctly uses the magnitude of `move_momentum` in places where a negative number would cause inconsistent results, like `fling_distance` and camera shake strength.

The code was there, implemented, called, but never actually fulfilled its own conditions -- except for the tank driving backwards at full speed, which can't be the only intended use.

If crashing wasn't desired, it would be better to remove it completely rather than leaving it in its bugged state.

Boots without issue. Crashing visual and fling effects now work for both forward and backward collisions.

# Changelog

:cl:
fix: vehicle crashes more consistently fling loose contents
/:cl:

### DIFF
--- a/code/modules/vehicles/multitile/multitile_movement.dm
+++ b/code/modules/vehicles/multitile/multitile_movement.dm
@@ -159,7 +159,7 @@
 
 	// Crashed with something that stopped us
 	if(!can_move)
-		move_momentum = Floor(move_momentum/2)
+		move_momentum = trunc(move_momentum/2)
 		update_next_move()
 		interior_crash_effect()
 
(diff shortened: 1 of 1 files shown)


AndroBetel:
will test merge for some time and then probably merge

PR #297, opened 2024-06-14
Coolifies the ARC

sunofang:
# About the pull request
Adds a few consoles to the ARC, and an ammo loader.

More useful ARC is better, this one how has overwatch consoles, and stuff.

Booted up, drove around in the ARC a bit, and made sure it all worked.

<details>

billede

billede

</details>

# Changelog

:cl:
add: Added small tacmap and overwatch console.
maptweak: Added overwatch to the ARC
/:cl:


DexterDude:
Not sure how I feel about this. I do get what you're going for, but the ARC is more intended to be an APC alternative for FORECON, hence why it has exactly 7 seats - this also allows GMs to use it in normal platoons by giving each squad a separate ARC. Since it's more FORECON oriented, and they don't really have a role that's meant to be on overwatch, the addition of the consoles really isn't all that useful.

However, I do think that the ideas here kinda have some merit. I definitely forgot to add the loader, which is my bad. 

My suggestions would be: Use the standard ARC layout but re-arrange it somewhat (since atm it isnt that good, too much is hidden behind chairs). Something like this:
[
thislayout
](url)

Use this piece of the sprite to represent the loader, maybe change it up, look a bit more like a loader (maybe with the conveyor belt sorta look the current one has)

BOXPILLAR

And you can turn this screen-looking thing into an overwatch console with a bit of a re-sprite, making the screens yellow/orange like the base overwatch console has. I think someone was showing that off recently and it looked quite nice, though I don't recall who.

thisspot

sunofang:
Good idea. I'll do that.

AndroBetel (changes requested):
i agree with verbs comment, 7 seats are minimum for ARC
maybe make the OW one a subtype?

AndroBetel (approved):

PR #303, opened 2024-06-22
Head Mounted Sight Has Camera

KoishiVibe:
# Summary
- The HMS now has a camera. Maybe. It might not work.
- The HMS now blocks equipping the M10 helmet, so that you don't have two camera feeds at once. 
- The HMS does not block equipping other head items.


KoishiVibe (commented):
fubar standby resetting

PR #305, opened 2024-06-22
Mk2 Super Dies

KoishiVibe:
# summary
- for all intents and purposes the mk2 should no longer be mentioned or appear. the mk2 is now a deep modification of the Mk1, the M41AE1. it doesn't make a huge amount of sense, but...


code/modules/projectiles/ammo_boxes/ammo_boxes.dm
```diff
@@ -257,7 +257,7 @@
 
 /obj/item/ammo_box/rounds
 	name = "\improper rifle ammunition box (10x24mm)"
-	desc = "A 10x24mm ammunition box. Used to refill M41A MK1, MK2, M4RA and M41AE2 HPR magazines. It comes with a leather strap allowing to wear it on the back."
+	desc = "A 10x24mm ammunition box. Used to refill M41A MK1, M4RA and M41AE2 HPR magazines. It comes with a leather strap allowing to wear it on the back."
```
private-tristan, line 260:
```suggestion
	desc = "A 10x24mm ammunition box. Used to refill M41A, M4RA and M41AE2 HPR magazines. It comes with a leather strap allowing to wear it on the back."
```

KoishiVibe: thanks yo

Doubleumc:
GMs have expressed a desire to keep the Mk2. Won't be moving forward with this.

PR #308, opened 2024-06-25
Removes Uniform Lock

KoishiVibe:
# Summary
- Removes uniform locks.
This prevents scenarios where someone might salvage protective equipment, but meta as "I need to have the right uniform to wear this". 
This also gives late down the line freedom of asthetics and mixing equipment, such as with perhaps, an insurgent faction...


AndroBetel:
yeah no, we shouldnt encourage marines changing their armor like that, morrows word

PR #309, opened 2024-06-25
Crates Are More Cratier

KoishiVibe:
# Summary (but actually when im not eepy)
- Crates now contain a wider array of items, and will contain _more_ items. 
- Crates from the middle of nowhere will not randomly contain marine equipment. 
- Attempt overall to pivot away from PVP13 survivor gameplay metafocus.
- More crate types, for use as crates that may be placed on the ship with contents that fit (planned)


AmoryBlaine:
This isn't really useful or adding much. I recommend you instead just create a series of crate loadouts for various locations and factions as a mapper resource. Also, the soap addition seems needless.

DexterDude (changes requested):
Here's my opinion on this.
You should say what you are changing and why you are changing it in the PR. Right now all I know is that some changes were made to crates to make them cratier because it's funny.
That doesn't exactly give me a good description of what you want to do nor what you are doing with the PR. Obviously I can see the files changed, but generally you should make it easy to view and go "oh these changes as a whole are neat" or "oh these changes as a whole are bad".

I have a few questions, namely, why did you make this PR and why did you make the specific changes? Without knowing this, all of the changes you have made seem kinda nonsensical. Why is clothing removed but sheets are added for example?

cm13-github:

KoishiVibe:
Guh. I was going to go somewhere with this and try to make different crate subtypes but this is already so far ebhind im not considering it worth the effort to bring up.

PR #323, opened 2024-07-05
Prop ropes! For all your ladder alternatives

AmoryBlaine:
Adds two different sets of rope sprites to the Almayer structure props that can be easily recoloured for use as alternate ladders

image


AndroBetel (changes requested):
thats kind of a half assed change tbh

either make rope a defined subtype of ladder, or add some variable that prevents ladders from updating the sprite on initialize

AmoryBlaine:
> thats kind of a half assed change tbh
> 
> either make rope a defined subtype of ladder, or add some variable that prevents ladders from updating the sprite on initialize
I don't know how to do that. As is all this does is make it so if you are so inclined, you can make ladders use rope sprites without a hassle and provides you with a colourable option. If someone else wants to add a subtype they're more than welcome to, I just don't know how.

PR #325, opened 2024-07-06
More cigarettes in pointbuy

Merrgear:
# About the pull request

Gives marines more pointbuy packets of cigarettes, because not every smoker likes lucky strikes. increased by 1-2 points than lucky strikes because they would be rarer, also adds gold zippo because some marine would be dumb enough to spend money on a gold plated zippo

More flavor for characters if John Marine is really an emerald smoker, subtle but something

# Changelog

:cl:
add: added more cigarette options in the loadout, along with a golden zippo
:cl:

### DIFF
--- a/code/modules/client/preferences_gear.dm
+++ b/code/modules/client/preferences_gear.dm
@@ -966,6 +966,36 @@ var/global/list/gear_datums_by_name = list()
 	display_name = "Pack Of Lucky Strikes"
 	path = /obj/item/storage/fancy/cigarettes/lucky_strikes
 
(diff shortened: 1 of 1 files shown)


Meatstuff882:
Merrgear I removed emeralds from the loadout menu specifically because morrow said "rare things should be rare" when I asked about making fancy cigs more common.

Merrgear:
> Merrgear I removed emeralds from the loadout menu specifically because morrow said "rare things should be rare" when I asked about making fancy cigs more common.

I mean, its a bit of a silly reason as he said, ima keep it up incase me and the other people who think so can change his mind about it

PR #339, opened 2024-07-23
Dropship roster expansion

DexterDude:
changes dropships

- makes midway green and adjusts its size, removing the side doors and changing its proportions
- does the same to the cyclone and also makes its turfs unique to it rather than using normandy
- adds the tripoli, which is essentially a tan midway
- adds the typhoon, a command dropship fitted with a surgical bed, TOC for the lieutenant, as well as gatling and railgun consoles that can be unlocked by GMs
- adds the tornado, a more transport oriented dropship capable of seating 17 marines, excluding the cockpit
- gives all dropships (except alamo/normandy) movie cockpits with tandem seating
- also gives all dropships unique structures for weapons, electronics, and everything else

- kinda changes the balance of the railgun console so that it's a bit less powerful
- changes the dropships on chances claim to reflect these changes
- probably other stuff i forgot


cm13-github:

PR #372, opened 2024-08-13
Load corpses into hypersleep chambers.

Ediblebomb:
Literally just read the title.


AndroBetel (changes requested):
can you make it so it doesnt cryo them and the corpse just sits inside forever

AndroBetel (approved):

PR #380, opened 2024-08-15
Removes Synthetic whitelist requirements for synth pref menu

KoishiVibe:
# summary
- removes synth prefs menu whitelist only display code
- does not add option to spawn them into an ERT, that'd for someone else to create
- maybe a requests system?
- this just allows someone to customize and set synth chars, but spawning them will still be ghost + special pick or opening the role up I think.


DexterDude:
cant speak to as if this works or not but I do think this is a great idea actually

KoishiVibe:
> cant speak to as if this works or not but I do think this is a great idea actually

ill do more extensive testing at some point
it was late night coding though so operability is ????

Doubleumc (approved):
Looks good to me.

PR #408, opened 2024-08-22
Limited Service/Out of Service Equipment Adjustments

Max-023:
# About the pull request
This pull request removes the shotgun scabbard and pouch from the vendors of the USCM and instead places them in the M37 boxes to keep consistent with the limited service nature the M37 has. Generic gas masks have also been removed from all prep vendors and replaced with service models, the generics now being considered 'civilian grade'. Civilian gas masks can still be selected from the preferences menu, however.

The new masks to replace are the M5 Standalone Gasmask and the PMK-63 Gasmask. The M5 Standalone is an alternative variant to the M5 Integrated for use with individuals who are not wearing or do not have access to the M10 Pattern helmet. The PMK-63 as well as its front line variant, the PMK-63N, are gas masks designed for UPP troops and act as the counterpart to the M5. The 63N more specifically is designed to attach to in-service helmets within the UPP Armed Forces.

The shotgun stuff is more of a minor thing that is just inconsistent with the M37s. The Marines don't need *that many* shotgun shell pouches when they usually only get enough shells to fill two and they don't need fifteen shotgun scabbards when they only get two shotguns.

The gas masks are good because the generic one tonally doesn't make sense to be both used by colonists as well as military personnel, especially personnel of **both sides**. The M5 Standalone is necessary to keep Marines aesthetically consistent and both PMKs are necessary to further separate the UPP into being its own faction rather than continue to exist using USCM gear to fill the gaps it has.

N/A

# Changelog

:cl:
add: Added PMK-63 and PMK-63N to UPP vendors
add: M5 Standalone to USCM regular and FORECON vendors
add: A shotgun scabbard and shotgun shell pouch to M37 box kits
del: Removed generic gas masks from squad prep vendors
del: Shotgun scabbards and shotgun shell pouches from USCM regular and FORECON vendors
tweak: Generic gas masks now cost only 1 point for those that still want to use them
tweak: Generic gas masks now are referred to as 'civilian grade' to denote their status as a civilian model
tweak: Altered the description of the M5 Integrated Gasmask to not be the haha funny jarhead PVP lore
/:cl:


code/modules/clothing/masks/gasmask.dm
```diff
@@ -31,6 +31,18 @@
 	icon_state = "kutjevo_respirator"
 	item_state = "kutjevo_respirator"
 
+/obj/item/clothing/mask/gas/m5
+	name = "\improper M5 Standalone Gasmask"
```
AndroBetel, line 35: no capital letters

code/game/machinery/vending/vendor_types/squad_prep/squad_prep.dm
```diff
@@ -117,7 +116,6 @@
 		list("First-Aid Pouch (Injectors)", round(scale * 15), /obj/item/storage/pouch/firstaid/full, VENDOR_ITEM_REGULAR),
 		list("Flare Pouch (Full)", round(scale * 15), /obj/item/storage/pouch/flare/full, VENDOR_ITEM_REGULAR),
 		list("Magazine Pouch", round(scale * 15), /obj/item/storage/pouch/magazine, VENDOR_ITEM_REGULAR),
-		list("Shotgun Shell Pouch", round(scale * 15), /obj/item/storage/pouch/shotgun, VENDOR_ITEM_REGULAR),
```
AndroBetel, line 120: with amory's PR we're getting multiple shell boxes, what if i want to put them all in my pockets? i see no reason for removing this pouch from the vendors

AndroBetel (changes requested):

code/game/machinery/vending/vendor_types/squad_prep/squad_prep.dm
Max-023: Given that this is a PR for the current state of the game, not for Amory's PR, I am willing to modify it to account for Amory's changes if they are fully merged and add a second pouch. The point of removing them from the vendors is because they are not standard service, so they should be lumped into the case with the weapon as well to fully reflect its use within the USCM.

code/modules/clothing/masks/gasmask.dm
Max-023: fixed

AndroBetel (approved):

PR #423, opened 2024-08-30
Muted Janitorial Carts and Cleaner bottles

AmoryBlaine:
- Blue and tan cleaner bottle icons and onmobs
- Blue and brown janitorial carts
Not the biggest fan of these replacements but also even less of a fan of the current stuff. Contrasts a lot with the environments they're in.
image
image


code/game/objects/structures/janicart.dm
```diff
@@ -16,6 +16,9 @@
 	var/obj/item/reagent_container/glass/bucket/janibucket/mybucket
 	var/signs = 0 //maximum capacity hardcoded below
 
+/obj/structure/janitorialcart
+	desc = "A brown janitorial cart that has space for water, mops, signs, and trash bags."
+	icon_state = "cart_brown"
```
Doubleumc, line 21: `/obj/structure/janitorialcart` is already defined as the blue cart. You're redefining the same thing here which overwrites that original definition, so *all* janitorial carts would be brown. Instead make a subtype, like this:
```suggestion
/obj/structure/janitorialcart/brown
	desc = "A brown janitorial cart that has space for water, mops, signs, and trash bags."
	icon_state = "cart_brown"
```

Doubleumc (changes requested):

Doubleumc (approved):

cm13-github:

cm13-github:

PR #436, opened 2024-09-05
Adds a 'slot nickname' field for easily ID'ing chara slots

BonniePandora:
# About the pull request

Adds a new field on the character setup screen with which to give the save slot a 'nickname' for easier identification at a glance, so you can tell what characters are for USCM, UPP, TWE, etc etc.

No more getting mildly confused about which chara is meant for which faction if you have a gagglefuck of them made up.

Tested extensively on a local instance, works without issue and as intended.
<details>

image
The final version of the code removed the (brackets) from either side of the slot nickname result on the setup page, but retains it on the load-slot window.

</details>

# Changelog

:cl:
qol: Adds a new 'nickname' for character slots
ui: adjusted the character setup menu to have a field for the above
/:cl:

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -117,6 +117,7 @@ var/const/MAX_SAVE_SLOTS = 20
 
 	//character preferences
 	var/real_name //our character's name
+	var/nickname //the nickname for the saveslot
 	var/be_random_name = FALSE //whether we are a random name every round
 	var/human_name_ban = FALSE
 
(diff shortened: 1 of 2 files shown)


Doubleumc (changes requested):
To avoid confusion, instead of calling it "nickname"/`nickname` call it something like "slot label"/`slot_label`.

Other than that, code good.

BonniePandora:
> To avoid confusion, instead of calling it "nickname"/`nickname` call it something like "slot label"/`slot_label`.
> 
> Other than that, code good.

Changed from nickname in all the visible stuff!

Doubleumc (approved):

PR #438, opened 2024-09-05
Weird Flavor Text & Item Naming Corrections

KoishiVibe:
# summary
- fuckin a metal riot shield heavy but effective


KoishiVibe (commented):
I can't the weird The shrapnel thing
maybe its an andro thing

PR #469, opened 2024-09-23
Xeno blood doesn't track

AmoryBlaine:
- Removes xeno blood overlays on mobs
- Removes xeno blood trailing on the floor


Doubleumc (approved):

Doubleumc:
Not going in this direction at this time. While there was some interest in changing up acid decals, there wasn't any interest in a full removal.

PR #481, opened 2024-09-27
Updates synth's headset tracking options

BonniePandora:
# About the pull request

Adjusts the available tracking options on the synthetic headset to be in-line with marine headsets, also changes the default tracking option selected from the CO to Platoon sergeant

As it stood, only the alpha squad lead & landing zone tracking options functioned on the synth headset, now this should allow them to track the LT too and remove some of the clutter on that menu at the same time.

Compiled fine, code largely copied from the standard headset sans the squad sergeant option.

# Changelog

:cl:
qol: USCM Synth headsets no longer cluttered up with all the PVP tracking options, now they can track PVE relevant options only, with the Platoon sergeant being the default option.
/:cl:


code/game/objects/items/devices/radio/headset.dm
```diff
@@ -610,18 +610,12 @@
...
-		"Executive Officer" = TRACKER_XO,
-		"Landing Zone" = TRACKER_LZ,
-		"Alpha SL" = TRACKER_ASL,
-		"Bravo SL" = TRACKER_BSL,
-		"Charlie SL" = TRACKER_CSL,
-		"Delta SL" = TRACKER_DSL,
-		"Echo SL" = TRACKER_ESL,
-		"Foxtrot SL" = TRACKER_FSL
+		"Platoon Commander" = TRACKER_PLTCO,
+		"Platoon Sergeant" = TRACKER_SL,
```
AndroBetel, line 617: this checks for user's assigned_squad, synths normally dont have one, so this option wont work

BonniePandora: Addressed, will track Alpha Squad Lead instead of the assigned_squad leader
I know that _that_ works for tracking the platoon sarge due to past experience

DexterDude:
Default should be platoon sergeant given they always deploy and synths usually do

AmoryBlaine:
You should just give them the standard headset

PR #483, opened 2024-09-28
MREs minus fortune cookie and water bottle

AmoryBlaine:
Water bottle being in there is stupid, fortune cookie was also weird. Food, cigs and matches are all you get now.


DexterDude:
fortune cookie is funny and ive seen some rp come from it

Doubleumc:
Not a direction we'd like to go at this time. GMs overwhelmingly preferred to keep fortune cookies and bottled water.

PR #503, opened 2024-10-12
Make Arterial Bleeds Disco Again

The32bitguy:
# About the pull request

This PRs flavours adds an intial, more severe stage of IB/Internal Bleeding called Arterial Bleeding:

- Arterial Bleeds are applied, and removed from limbs exactly as before.
- They **passively** remove blood at 50% the rate of IB every tick.
- They **actively** remove blood at a 10% chance every tick by making a blood spray.
- The blood spray has an exception to bypass the 10% chance on the first tick after the bleeding effect is given to a limb.
- This blood spray is a series of blood effects made across turfs at a distance proportional to the remaining blood volume of the human. It also makes a sound. Each time it happens, it removes a moderate amount of blood inversely proportional to the remaining blood volume of the human.
- The chatlog states which limb has this bleeding effect.
- The blood spray stops early and blurs vision if it hits a human.
- The active part can be stopped by applying gauze or a trauma kit to the limb, which uses one stack and takes priority over the treatment of the wound on the limb, providing the limb has the bleeding effect. Otherwise the gauze/trauma kit is applied as normal.
- There is a chance every time the applied gauze prevents the active part for it to expire, necessitating another application of gauze. When it expires it leaves a used bandage prop on the ground.
- Once the active part is stopped the scanner display changes from [Arterial Bleeding!] to [Internal Bleeding], and effectively acts identical to old IB except at a 50% rate.
- The chatlog information shown when examining your character on help intent now shows if a limb has this bleeding effect and rather or not it was mitigated with gauze/trauma kit.

Things to do:
- Make the blood spray check for pass flags and not pass through objects like windows.
- ~~Implicit src~~
Sound is from https://freesound.org/people/kyles/sounds/453769/

Internal bleeding is not very interactive in the game. There is no way to tell when your character has gotten one except for [one exception](https://github.com/cmss13-devs/cmss13-pve/blob/0a7b111584bccaf199364e4177befc746c18a489/code/modules/organs/limbs.dm#L453) which rarely occurs because of the prevalence of armour that characters wear. To check if a character has one requires a health analyser or waiting for the effects of blood loss to become apparent, and therefore crippling.

Giving a graphical, audible and chatlog indication will offset the slight additional mechanical complexity by making it relatively easy to understand what is occurring(big bleed, this needs fixing!). This also allows marines to gives meaningful first aid that isn't performing the job of the corpsman but with less reagents, and reduces the dependency on the health analyser. 

<details>

image

</details>

# Changelog
:cl:
add: Internal Bleeding now instead starts as Arterial Bleeding, that sprays blood with a visual, audible and chatlog indication
qol: Arterial Bleeds now show on help-intent self examine.
balance: Arterial Bleeds can have the spraying component mitigated by applying gauze or a trauma kit, turning them into a slower version of the old Internal Bleeding effect.
balance: Iron Pills have 7.5u of sugar in them.
balance: The IB treatment step now takes 7.5 seconds instead of 5 seconds.
soundadd: The sound for the above blood spray.
imageadd: The icons for each possible direction of the blood spray, to avoid overlays inheriting the turfs direction for their icon, due to blood optimization.
/:cl:


code/datums/effects/bleeding.dm
```diff
@@ -113,7 +115,13 @@
 				return FALSE
 
 	blood_loss = max(blood_loss, 0) // Bleeding shouldn't give extra blood even if its only 1 tick
-	affected_mob.blood_volume = max(affected_mob.blood_volume - blood_loss, 0)
+	affected_mob.blood_volume = max(affected_mob.blood_volume - blood_loss*0.5, 0) //
+	if(prob(10) || src.show_spray_immediately)
```
AndroBetel, line 119: no need to use src. here and further

code/modules/mob/living/blood.dm
```diff
@@ -310,3 +309,34 @@
 	var/obj/effect/decal/cleanable/blood/oil/O = locate() in T.contents
 	if(!O)
 		O = new(T)
+
+/mob/living/carbon/human/proc/spray_blood(turf/T, spray_dir, limb)
+	var/obj/limb/O = limb
+	src.visible_message(\
```
AndroBetel, line 315: no need for src. here and further

code/modules/mob/living/blood.dm
```diff
@@ -310,3 +309,34 @@
...
+	for(var/i = 1 to ceil(src.blood_volume/150))
+		T = get_step(T, spray_dir)
+		if(T.density)
+			break
+		for(var/A in T)
+			var/mob/living/carbon/human/sprayed_with_blood = A
+			if(ishuman(sprayed_with_blood))
+				if(!sprayed_with_blood.body_position == LYING_DOWN)
+					sprayed_with_blood.add_mob_blood(src)
+					//if(rand_zone() == "head")
```
AndroBetel, line 328: redundant comment

code/game/objects/items/stacks/medical.dm
```diff
@@ -290,6 +301,28 @@
 			var/message = SPAN_WARNING("[user == M ? "You don't" : "[M] doesn't"] have \a [limb]!")
 			to_chat(user, message)
 			return
+		//Tourniquet functionality
+		for(var/datum/effects/bleeding/internal/I in affecting.bleeding_effects_list)
```
AndroBetel, line 305: this should be done by a different item (a proper tourniqet)

code/modules/organs/limbs.dm
```diff
@@ -1261,6 +1261,22 @@ treat_grafted var tells it to apply to grafted but unsalved wounds, for burn kit
...
+		user.visible_message(SPAN_WARNING("[user] fumbles with [S]"), SPAN_WARNING("You fumble with [S]..."))
+		time_to_take = 10 SECONDS
+	if(do_after(user, time_to_take * user.get_skill_duration_multiplier(SKILL_MEDICAL), INTERRUPT_NO_NEEDHAND, BUSY_ICON_FRIENDLY, target, INTERRUPT_MOVED, BUSY_ICON_MEDICAL))
+		var/possessive = "[user == target ? "your" : "\the [target]'s"]"
+		var/possessive_their = "[user == target ? user.gender == MALE ? "his" : "her" : "\the [target]'s"]"
+		user.affected_message(target,
+		SPAN_HELPFUL("You tighten the <b>splint</b> on [possessive] <b>[display_name]</b>, <b>resticting the blood flow to the damaged artery.</b>"),
+		SPAN_HELPFUL("[user] tightens the <b>splint</b> on your <b>[display_name]</b>, <b>resticting the blood flow to the damaged artery.</b>"),
+		SPAN_NOTICE("[user] tightens the <b>splint</b> on [possessive_their] [display_name], <b>resticting the blood flow to the damaged artery.</b>"))
+		I.has_been_bandaged = TRUE
```
AndroBetel, line 1276: having to pack it with gauze seems like a move in any given situation, tq should have some advantages

AndroBetel (changes requested):
see above

code/datums/effects/bleeding.dm
```diff
@@ -113,7 +115,13 @@
...
+		if(!src.has_been_bandaged) //If Arterial has been packed, only remove blood passively every tick
+			show_spray_immediately = FALSE
+			affected_mob.spray_blood(get_turf(affected_mob), pick(GLOB.alldirs), src.limb)
+			affected_mob.blood_volume = max(affected_mob.blood_volume - blood_loss*20*(affected_mob.blood_volume/BLOOD_VOLUME_NORMAL), 0)
```
AndroBetel, line 123: `20` should be a define, else its just some random arbitrary number

AndroBetel (dismissed):

code/modules/organs/limbs.dm
The32bitguy: It seems to be a bind between making it a medical item that would be carried by a non-medic and making it a dedicated tq item, and how they would differ in effectiveness. Maybe make it gauze for now.

code/game/objects/items/stacks/medical.dm
The32bitguy: Changed to using gauze https://github.com/cmss13-devs/cmss13-pve/blob/9f5c4aca39d08601830572fbbb91e0f2a68a4b92/code/game/objects/items/stacks/medical.dm#L82-L85

Changed to using trauma kit https://github.com/cmss13-devs/cmss13-pve/blob/9f5c4aca39d08601830572fbbb91e0f2a68a4b92/code/game/objects/items/stacks/medical.dm#L196-L199

code/datums/effects/bleeding.dm
The32bitguy: https://github.com/cmss13-devs/cmss13-pve/blob/9f5c4aca39d08601830572fbbb91e0f2a68a4b92/code/datums/effects/bleeding.dm#L5

cm13-github:

cm13-github:

cm13-github:
## Test merge deployment history:

<details><summary>Test Merge Updated @ 03/23/2025 09:05:10 +00:00:</summary>

##### Server Instance
PVECM: Green

##### Merged By
Drathek

##### Revision
Origin: 1172d1f1588b01be620a11e22ea32db183fc2560
Pull Request: cdd07ebee15a056d0978cbf8260201fefe39a283
Server: 3869c77485a8e03a741435052c3c16ca233f3258
</details>

BonniePandora (approved):
Code lgtm, changes have been addressed & the PR's commonly used in test-merges

PR #515, opened 2024-10-22
Bayonets no longer mount on M41A, UGL no longer removable

AmoryBlaine:
For purposes of lore and anti-powergame, the UGL is no longer removable from the M41A, nor do bayonets fit on the M41A. No more stabbing xenomorphs and resin, it makes no sense and is gamey.


Doubleumc (approved):
Code good.

PR #537, opened 2024-11-04
Doors Updated

AmoryBlaine:
- Unifies the hazard strip on white generic doors with other doors
- Unifies the 'medical' doors by removing the cross so they can be used more liberally. Seemed needless to have two crosses on every set of doors when we have semiotics.


icons/obj/structures/doors/medidoor.dmi
Doubleumc, line 1: This file has been modified, but there are no sprite changes. Likely you modified something in here to test, but then undid it. That may have changed some metadata in the `.dmi`. Can get a fresh copy from the master.

Doubleumc (changes requested):
Other than that, code good.

icons/obj/structures/doors/medidoor.dmi
AmoryBlaine: Yeah, I had a feeling that'd happen, I replaced then created a new one. Annoying, it's the same thing that happened back when with the UPP one.

Doubleumc (approved):

PR #539, opened 2024-11-06
Wirecutter buff

AmoryBlaine:
Fences now have 6500 health. Use wirecutters to bypass.

### DIFF
--- a/code/game/objects/structures/fence.dm
+++ b/code/game/objects/structures/fence.dm
@@ -8,9 +8,9 @@
 	anchored = TRUE
 	layer = WINDOW_LAYER
 	flags_atom = FPRINT
-	health = 50
+	health = 6500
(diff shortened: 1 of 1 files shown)


DexterDude:
if theres an alternative to this that prevents marines from shooting them down/knifing them instead of a flat health buff that would be preferred

BonniePandora:
> if theres an alternative to this that prevents marines from shooting them down/knifing them instead of a flat health buff that would be preferred

Line 63 to 65 in the .dm being commented out/removed ought to achieve that, that's where the checks for items hitting the fences is.
Or if thrown items _eventually_ breaking the fence is still wanted, adjustment of how tforce on line 65 is calculated could be done (divide the I.throwforce value by five or something?)

PR #550, opened 2024-11-14
FORECONless Chance

AmoryBlaine:
Makes the default version of Chance into a version without FORECON, adds a new FORECON one


Doubleumc (approved):
Code good.

Would prefer to see this as a nightmare insert rather than a whole "new" map, but don't think the insert system can handle all that.

PR #586, opened 2024-12-03
Atomized CLF Removal

KoishiVibe:
# Summary
- Makes the CLF a more generic Insurgents, and removes their special radio tags over radio.
- Removes all mentions of CLF.

# Justification
- It feels incredibly gamey to have a face faction, a named, DISTINCT, face faction, who is supposed to be an insurgency, a nameless shadow organization.

# Changelog
:cl:
add: Guerrilla and Insurgent in place of CLF.
add: Colonist doctors now have radio roles. 
del: CLF mentions almost entirely removed.
balance: Insurgents are now entirely regular, down to the ID card giving them normal jobs.
/:cl:


Doubleumc (changes requested):
Two points:
 - You're calling the former CLF "guerrilla", "insurgent", "rebel", and "irregular" interchangeably. Its not immediately clear these different words are referring to the same faction. Its okay to use varied words in lore/descriptions, but for titles or names of stuff pick just one and stick with it.
 - You've removed the distinguishing part of the job names. I get that blurring the lines between combatants and non-combatants is kind of an insurgency's "thing" but that's an entirely different discussion if we want that in our spaceman game. Re-add in your descriptor of choice.

KoishiVibe:
> Two points:
> 
>     * You're calling the former CLF "guerrilla", "insurgent", "rebel", and "irregular" interchangeably. Its not immediately clear these different words are referring to the same faction. Its okay to use varied words in lore/descriptions, but for titles or names of stuff pick just one and stick with it.
> 
>     * You've removed the distinguishing part of the job names. I get that blurring the lines between combatants and non-combatants is kind of an insurgency's "thing" but that's an entirely different discussion if we want that in our spaceman game. Re-add in your descriptor of choice.

On the first, I've unified it to insurgency since guerrilla implies being an arm of some state. Insurgent is actual rebellion type stuff. 

On the second... I'm going to wait on that, I think.

Doubleumc:
Just noting here that I'm satisfied with the unified faction names. Will wait for the job names.

PR #617, opened 2024-12-20
Allows Chances Claim [Forecon] to be selected

private-tristan:
# About the pull request
When   #550 was merged, it didn't actually add anything other than the map file for chances claim forecon variation, meaning that the only way to select it was to override the map (very inconvenient).

Also changes a bit of text for FORECONLESS chance's so that it doesn't mention forecon, only the colony

GMs should be able to pick forecon chances claim without needing to manually override the map

image

# Changelog

:cl:
fix: Chances Claim [Forecon] can now be properly selected by GMs
/:cl:

### DIFF
--- a/maps/lv522_chances_claim_forecon.json
+++ b/maps/lv522_chances_claim_forecon.json
@@ -1,8 +1,8 @@
 {
-	"map_name": "LV-522 Chance's Claim",
+	"map_name": "LV-522 Chance's Claim Forecon",
     "short_name": "Chance's Claim",
-	"map_path": "map_files/LV522_Chances_Claim",
-	"map_file": "LV522_Chances_Claim.dmm",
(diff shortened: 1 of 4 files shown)


maps/lv522_chances_claim_forecon.json
```diff
@@ -0,0 +1,38 @@
+{
+	"map_name": "LV-522 Chance's Claim Forecon",
```
DexterDude, line 2: Just leave this without the forecon as I believe it shows up as the map name in game and I think that'd be a bit awkward.

private-tristan: But that would result in 2 maps named Lv-522 Chance's Claim, and afaik the map picker uses the map name, resulting in 2 names, is there any way you know to avoid this?

PR #637, opened 2024-12-30
Fix Lizards, Cats and Parrots no longer push you around

AmoryBlaine:
# About the pull request

No more pushing by certain simple mobs

Annoying and clearly not intended.

<details>

</details>

# Changelog

:cl: AmoryBlaine

fix: pushing by certain simple mobs
/:cl:


code/modules/mob/living/simple_animal/friendly/cat.dm
```diff
@@ -16,6 +16,8 @@
 	response_help  = "pets"
 	response_disarm = "gently pushes aside"
 	response_harm   = "kicks"
+	status_flags = list(CANSTUN, CANKNOCKDOWN, CANKNOCKOUT, CANDAZE, CANSLOW)
```
Doubleumc, line 19:
```suggestion
	status_flags = parent_type::status_flags & ~CANPUSH
```
`status_flags` takes bit toggles, so would be something like `CANSTUN|CANKNOCKDOWN|CAN...` instead of a list.

Instead of redefining `status_flags`, all we want to do is remove `CANPUSH`.
The default mob `status_flags` is... `DEFAULT_MOB_STATUS_FLAGS`. Defined here:
https://github.com/cmss13-devs/cmss13-pve/blob/04ffd839364b07dd0ea2d36ae57bb4cc2468705b/code/__DEFINES/mob.dm#L1
`STATUS_FLAGS_DEBILITATE` is defined here:
https://github.com/cmss13-devs/cmss13-pve/blob/04ffd839364b07dd0ea2d36ae57bb4cc2468705b/code/__DEFINES/mobs.dm#L105
So ultimately the default mob `status_flags` is equivalent to `CANKNOCKDOWN|CANPUSH|CANSTUN|CANKNOCKOUT|CANDAZE|CANSLOW|CANROOT`

We just want to remove `CANPUSH` from that. Can do that by using the scope operator to reference its existing value rather than redefining it: [https://www.byond.com/docs/ref/#/operator/::](https://www.byond.com/docs/ref/#/operator/::)
`parent_type::status_flags` is taking the existing var from its parent, the default listed above.
`~CANPUSH` is taking the boolean inverse of `CANPUSH`, meaning the `CANPUSH` flag is FALSE and all other flags are TRUE.
`parent_type::status_flags & ~CANPUSH` toggles OFF any bits that are false, so toggles off the `CANPUSH` flag of the parent's flags.

code/modules/mob/living/simple_animal/friendly/lizard.dm
```diff
@@ -19,3 +19,5 @@
 	black_market_value = 50
 	dead_black_market_value = 0
 	squeeze_under = TRUE
+	status_flags = list(CANSTUN, CANKNOCKDOWN, CANKNOCKOUT, CANDAZE, CANSLOW)
```
Doubleumc, line 22:
```suggestion
	status_flags = parent_type::status_flags & ~CANPUSH
```

code/modules/mob/living/simple_animal/parrot.dm
```diff
@@ -33,7 +33,7 @@
 	icon_living = "parrot_fly"
 	icon_dead = "parrot_dead"
 	mob_size = MOB_SIZE_SMALL
-
+	status_flags = list(CANSTUN, CANKNOCKDOWN, CANKNOCKOUT, CANDAZE, CANSLOW)
```
Doubleumc, line 36:
```suggestion
	status_flags = parent_type::status_flags & ~CANPUSH
```

code/modules/mob/living/simple_animal/friendly/cat.dm
```diff
@@ -16,6 +16,8 @@
...
+	layer = 3.5
```
Doubleumc, line 20: Why is the layer being changed to 3.5?

code/modules/mob/living/simple_animal/friendly/lizard.dm
```diff
@@ -19,3 +19,5 @@
...
+	layer = 3.5
```
Doubleumc, line 23: Why is the layer being changed to 3.5?

Doubleumc (dismissed):

code/modules/mob/living/simple_animal/friendly/cat.dm
AmoryBlaine: I have much to learn

code/modules/mob/living/simple_animal/friendly/cat.dm
AmoryBlaine: Makes them easier to pick up on the same tile

code/modules/mob/living/simple_animal/friendly/lizard.dm
AmoryBlaine: Makes them easier to pick up on the same tile

PR #647, opened 2025-01-05
Flavour Text and Character Records can now be Longer.

The32bitguy:
# About the pull request

Raises the character limit of flavour texts to the current limit of character records, from 1024 to 3072.
Raises the character limit of character records, from 3072 to 10240.
The input box of the Set Pose box is no longer one line long. It is instead now much larger(see screenshots).

I noticed two characters today that had their flavour texts cut off early due to the limit. Also the character record limits are rather short considering the context they are written for, sitting down to read paperwork someone can invest time outside of the game into writing.

image

# Changelog

:cl:
ui: Raised character limit of flavour texts from 1024 to 3072.
ui: Raised character limit of character records from 3072 to 10240.
ui: The input box of the Set Pose IC verb is larger.
/:cl:


Doubleumc (changes requested):
Would prefer you make new defines and use them as opposed to repurposing existing defines. Something like `#define MAX_RECORDS_MESSAGE_LEN 2048` or `#define MAX_DESC_MESSAGE_LEN 2048`
https://github.com/cmss13-devs/cmss13-pve/blob/04ffd839364b07dd0ea2d36ae57bb4cc2468705b/code/__DEFINES/text.dm#L22-L26

Just as a note, am hesitant of this PR considering the warning just above said defines:
https://github.com/cmss13-devs/cmss13-pve/blob/04ffd839364b07dd0ea2d36ae57bb4cc2468705b/code/__DEFINES/text.dm#L21

The32bitguy:
> Would prefer you make new defines and use them as opposed to repurposing existing defines. Something like `#define MAX_RECORDS_MESSAGE_LEN 2048` or `#define MAX_DESC_MESSAGE_LEN 2048`
> 
Yes.
> 
> Just as a note, am hesitant of this PR considering the warning just above said defines:
> 
I will try my best to DOS my local, and once this is test merged, the production server. If it an issue then I can leave the character record limit unchanged and instead split the 'load' by letting character slots fill the other existing fields in the records UI. 

A full redo of the UI is for another PR.

Doubleumc (approved):

PR #545, opened 2024-11-10
Dropship Airlocks

Warfan1815:
I hate writing PR descriptions after so much effort making it.
Do not merge until I've added the sound effects and the bugs have been fixed

Here's a video of it in action (w/o sound effects and with some bugs):

PR #651, opened 2025-01-05
Removes base area lighting from the dropships.

Ediblebomb:
# About the pull request
Base area lighting in a dropship with a closed roof and _lights on the map_ doesn't really have a reason to exist, this also adds a light to the cockpit so it's not pitch black in there.
Base area lighting can look pretty silly (see below image), and the dropships already have lights mapped.
<details>
image
image
</details>

# Changelog
:cl: Ediblebomb
maptweak: Dropships no longer have base area lighting.
/:cl:

### DIFF
--- a/code/game/area/Sulaco.dm
+++ b/code/game/area/Sulaco.dm
@@ -121,6 +121,7 @@
 	soundscape_interval = 30
 	is_landing_zone = TRUE
 	ceiling = CEILING_REINFORCED_METAL
+	base_lighting_alpha = 0
 
 /area/shuttle/midway/Enter(atom/movable/O, atom/oldloc)
 	if(istype(O, /obj/structure/barricade))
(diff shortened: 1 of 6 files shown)


harryob:
this exists because they get fullbright in transit anyway and it looks pretty dumb toggling between soft ambient lights to harsh lights for no reason when it launches

Warfan1815:
> > this exists because they get fullbright in transit anyway and it looks pretty dumb toggling between softer ambient lights to harsh lights for no reason when it launches
> 
> It doesn't look too bad to me. The dropships have enough lights that there aren't really shadows anywhere, though the soft shadows on the walls disappearing does look a little odd, and the cockpit light being red is... bad, I'll probably change it.

This problem has already been, more or less, fixed in #545 (which is nearing merge) . I can't remember specifically how I did it, but it looks alright transitioning between space and ground.

PR #671, opened 2025-01-16
Remove Non-Crit OD Coffee Jitter

The32bitguy:
# About the pull request

Coffee no longer causes jitter unless it is a critical OD.

1.05 Cups of Coffee should not cause the same effect as chestbursting, neurotoxin, etc.
# Changelog

:cl:
del: Coffee only makes a human jittery at a critical-OD threshold, not from a consecutive > 20 units.
/:cl:


sippykot:
sovl removal

PR #688, opened 2025-01-20
Fixes inconsistences with pronouns in emotes

The32bitguy:
# About the pull request

Stops the words "their/them/they", and other pronouns being swapped in the _**chatlog**_ with the equivalent of the character's pronoun/gender when they do a /me.
Langchat now shows the characters pronouns for _**predefined**_ emotes, same as the chatlog.

It is unnecessary and confusing to not be able to use these words in an emote, and to have your emote subtly changed.

CUSTOM EMOTE TEXT:
image
BEFORE:
Screenshot 2025-01-20 020413
image
AFTER:
Screenshot 2025-01-20 020413
image
----
PREDEFINDED EMOTE TEXT:
image

BEFORE:
image
image

AFTER:
image
image

<details>

</details>

# Changelog

:cl:
fix: Pronouns are no longer force replaced in the chatlog output of a Me verb with the character's equivalent. No more "their" --> "her".
fix: Overhead chat now uses the character's pronouns for predefined emotes, like the chatlog already does.
/:cl:

### DIFF
--- a/code/datums/emotes.dm
+++ b/code/datums/emotes.dm
@@ -18,6 +18,10 @@
 	var/message = ""
 	///Message displayed for audible emotes if someone's deaf. Only use this if key_third_person can't be acceptably used.
 	///ie. if someone says *medic and someone's deaf, they'd receive [X] calls for a medic! silently.,  which is not ideal.
+
+	/// Message after any alterations, like pronoun replacement for predefined emotes
+	var/msg =""
+
 	var/alt_message
 	/// Message with %t at the end to allow adding params to the message, like for mobs doing an emote relatively to something else.
 	var/message_param = ""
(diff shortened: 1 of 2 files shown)


Zonespace27 (changes requested):
This is not the right way to do this. This also affects all emotes (*shakehead for example) that are not manually entered. Additionally, just remove code instead of commenting it out.

code/datums/emotes.dm
```diff
@@ -18,6 +18,10 @@
 	var/message = ""
 	///Message displayed for audible emotes if someone's deaf. Only use this if key_third_person can't be acceptably used.
 	///ie. if someone says *medic and someone's deaf, they'd receive [X] calls for a medic! silently.,  which is not ideal.
+
+	/// Message after any alterations, like pronoun replacement for predefined emotes
+	var/msg =""
```
Zonespace27, line 23:
```suggestion
	var/msg = ""
```

Zonespace27 (changes requested):
Are emotes not singletons? This approach could cause issues

code/datums/emotes.dm
The32bitguy: What are the issues?

PR #704, opened 2025-01-24
Shrapnel Nerfs & Marine Armor Buffs

KoishiVibe:
# Summary
- Reduces the penetration of most shrapnel from AP tier 4, or P20, to negative armor piercing values. 
- Breaching charges and BFAB are unaffected; both have a reason to have their penetration so high. One is for gameplay reasons, the other is for the fact that the shrapnel grenade is creating extremely dense clouds of shrapnel that attack gaps in body armor.
# Changelog
:cl:
balance: Shrapnel of grenades across the board has reduced penetration and hence reduced effect against armored targets., instead of uh. Having P20 and just ignoring most armor???
/:cl:


KoishiVibe:
tempselfclose

PR #942, opened 2025-05-06
Remove lrp storage skill checks

sippykot:
# About the pull request

removes skillchecks from medkits, engineer kits and the pouches

the whole thing just exists as a gamey balance thing for pvp
you don't need special training to put a box in your bag or to put a battery in a pouch

# Changelog

:cl:
del: skill checks for putting medkits and engineer kits in backpacks
del: skill checks for using medkit and engineer kit pouches
/:cl:


sippykot:
i think the pill bottle skill checks should also go, but thats for another pr

PR #964, opened 2025-05-18
WY Rapid Response - The Heavy Tactical Officer

CactusMouth:
"Sir, hostiles are at the gate and our officers have been pushed back!"
**"Where the hell is our QRF?!"**

# About the pull request

When the PMCs are too far to respond and high value targets are under threat, the Heavy Tactical Officer responds as the bulwark against the horrors that strike out against Weyland Yutani.

Adds the Heavy Tactical Officers, armed with AP Viper SMGs, pistols, breaching charges, gas masks and up-armored ballistic armor, they are equipped to act as a QRF to any WY site under threat.

Also includes the Full Ballistic Vest, which is a slightly modified standard vest to take higher rounds.

WY Guards are generally pretty easy to deal with. Vipers are weak compared to other guns, and are underequipped to deal with most players/hostiles. 

These guards pose a bit more of a challenge, and should be used by GMs to ramp up the threat a bit more, but not enough to warrant PMCs. Gas masks restrict the use of chemical warfare to just gas out the hostiles and the stronger armor means that the players would have to be careful not to just stand out in the open and burst fire them, else risk AP returned fire.

The AP Vipers aren't honestly horrible to deal with though, so they should be fine but still.

Screenshot 2025-05-18 094135
Heavy Response Officer (Left) next to a standard Tactical Response Officer (right).

# Changelog

:cl:
add: Added Heavy Response Officer + Full ballistic vest.
code: Added the Heavy Response Officer
imageadd: Added the new armour set, Full Ballistic Armor
/:cl:


cm13-github:

PR #1062, opened 2025-06-26
Readds Autodoc Automatic Mode

Ediblebomb:
# About the pull request
You can now use autodocs regardless of surgical skill. Only on automatic mode though.
Autodocs being limited to medic, secsarge, and platco does not provide any real gameplay benefit other than screwing people over after the OP because those roles aren't present. An _autodoc_ should not require advanced training (you can literally do surgery yourself with that skill level) to utilize. If you can use a body scanner, you can use an autodoc. Also automatic mode is cool.
<details>

image

</details>

# Changelog

:cl:
add: Readds autodoc automatic mode
qol: You can now use autodocs regardless of surgical skill, however, you can only use automatic mode if you don't meet the skill requirements
/:cl:


mesnomer:
If the issue is post-op surgery, then why are we not simply swapping out the auto docs on the ship maps for the skilless type?

This is a balance swing for groundside medical -- one that you're not acknowledging in your rationale. I'd encourage you to make a case for that or to rethink your approach.

As for IC justification for how skill is required:

Modern medical devices take a lot of training to use correctly. Even largely automated ones require specialists to operate or at least supervise. There's a lot of nuance to the viability and approach of medical procedures based on the patient and on outside circumstances. Having a grunt that doesn't know the mechanics of a human heart or how to even tie a suture sign off on an open-heart surgery is incredibly silly. Especially when it's a machine in some random colony.

BonniePandora (changes requested):
Skillless ones for shipside please. There's two shipmap PRs currently, one of which can ignore the change entirely (`RMC & surg skill` moment)
In total, four ship-maps on your end to be updated, that isn't a lot. Chapev, rover, garrow, w.e. the default PMC one is called.
The only surgery which doesn't pop up on a healthy individual shoved inside is foreign object removal, which only crops up when the occupant has certain conditions met (namely having something implanted in them), if you want to make it so that crops up as standard on them normally then by all means, should stop the `magical anatomical assessment` stuff.

BonniePandora (changes requested):
There's some weird quirkiness (the surgery list not showing) with the menu & opening it with changed surg skill levels (0 -> 1 -> 2 -> 3) whilst someone's already *in* the thing, but that's honestly super niche and unlikely to be a problem. I do want to see the speed slowed further still, 2.5 or 3, so there's an actual **reason** to not go over the corpsman's helmet and toss people in the autodocs willy-nilly, the current speed of 1,5 doesn't feel noticeably longer than the usual speed.

BonniePandora (approved):

PR #1162, opened 2025-07-26
"Signal Tracker" for GM Objectives

Yorkoh:
# About the pull request

I'm just the idea person for this. @The32bitguy actually created the code block. https://discord.com/channels/1160427683151368192/1164728861766975568/1398472945751556126

The Signal Tracker is an item a GM can spawn to create tactile ways for players to track the objectives they mark via the GM Panel. Set an objective (mob, turf or object) and name it. The Signal Tracker can individually track these Objectives (but won't see colour, just the names). It doesn't have a battery life, unlike the PDT/L.

GMs can introduce this item as part of the gameplay loop: come up with a reason for these objectives to be tagged during a round so that players can track them (e.g. hack into a local network to find the hidden research site entrance, use a crew monitor to 'tag' important HVTs that are on the move). It also allows for objective tracking on large maps where GM marked objectives are not easily discerned on the tacmap (e.g the map is too big and stretches out, this aids in landnav). 

tested locally

# Changelog

:cl:
add: signal tracker TGUI item
add: signal tracker radar code
/:cl:


BonniePandora (approved):
Simple, elegant, useful.

PR #1168, opened 2025-07-28
More pistol stat tweaks

BonniePandora:
# About the pull request

T73 comparable to being UPP version of M1911, slightly less damager per hit but more rounds per mag. M1911 damage tweaked up slightly to mitigate the negative AP malus.
.38 Revolver no longer has stats designed by a crackhead, about on-par with the normal .357 rounds but a bit spicier
.44 automag took a damage reduction so it isn't _just_ as good as the M44 but with more rounds
NP92 now properly loads 9x18mm rounds in the magazine, rather than cramming 9x19 in there somehow

Should make pistols feel more fun to use without then overcooking hAI that come with a sidearm alone

Tested locally, nothing broken

# Changelog

:cl:
balance: T73 and M1911 damage upped slightly
balance: .38 revolver and .44 automag have had damage lowered
balance: .38 revolver no longer has a chance to send rounds careening off into the next post-code due to accuracy debuffs
fix: NP92 now chambers the advertised 9x18mm, rather than 9x19mm
/:cl:


code/modules/projectiles/guns/pistols.dm
```diff
@@ -305,6 +305,7 @@
 	set_fire_delay(FIRE_DELAY_TIER_6)
 	recoil = RECOIL_AMOUNT_TIER_5
 	recoil_unwielded = RECOIL_AMOUNT_TIER_3
+	damage_mult = BASE_BULLET_DAMAGE_MULT - BULLET_DAMAGE_MULT_TIER_3
```
AndroBetel, line 308: why is .44 magnum less damaging that .45

AndroBetel (changes requested):

code/modules/projectiles/guns/pistols.dm
AndroBetel: nvm

AndroBetel (approved):

